PR #1, opened 2016-08-12
TWILL-182, prefix jar names with the source path in the bundle if the…

serranom:
… name is not unique.  added test


hsaputra:
Could we build the `samename.jar` file as part of build system rather than checking it to source repo?

serranom:
Sure. That is a good idea.  I tried doing it in the test code but that
complicated the classpath such that the test was too hard to write.  I'm
not in this week but will update the pr next week.

twill-core/src/main/java/org/apache/twill/internal/ApplicationBundler.java
```diff
@@ -215,8 +215,19 @@ public boolean accept(String className, URL classUrl, URL classPathUrl) {
   private void putEntry(String className, URL classUrl, URL classPathUrl, Set<String> entries, JarOutputStream jarOut) {
     String classPath = classPathUrl.getFile();
     if (classPath.endsWith(".jar")) {
+      String entryName = classPath.substring(classPath.lastIndexOf('/') + 1);
+      // need unique name or else we lose classes (TWILL-181)
+      if (entries.contains(SUBDIR_LIB + entryName)) {
+        String[] parts = classPath.split("/");
```
chtyim: I don't quite understand this part of the logic. Is it trying to use some parent directory name as the entryName? Seems like it would lost the original name. Also, if the jar is already at the root directory, then it will be using the original "entryName", which would still result in name conflict? I would suggest simply use package name (derived from `className`) and a random uuid to make it unique if there is conflict. Something like:

```java
String entryName = classPath.substring(classPath.lastIndexOf('/') + 1);
if (entries.contains(SUBDIR_LIB + entryName)) {
  entryName = entryName.substring(0, entryName.lastIndexOf('.')) + "-" + className.substring(0, className.lastIndexOf('.')) + "-" + UUID.randomUUID() + ".jar";
}
```

chtyim:
+1 for @hsaputra idea of not checking in the binary jar. To not mess with the classpath, you can simply generate an empty class using ASM and write it to a jar (shouldn't be too complicate to use `ClassWriter` to do so, probably about 20 lines of code). I can help with that if needed :)

chtyim:
Actually you don't need to use ASM. You can just create two jars, one contains `org/apache/twill/internal/utils/Class1.class`, the other has `org/apache/twill/internal/utils/Class2.class`. Then create a `URLClassLoader` from that two jars, with `null` as the parent. Then load `Class1` and `Class2` from that classloader and provide the both classes to the `ApplicationBundler.createBundle` method.

twill-core/src/main/java/org/apache/twill/internal/ApplicationBundler.java
serranom: In this context, we know the classPath urls are all unique because they come from a set.  What this logic does is to prepending the entry name with enough elements of this path to make it unique.  I presumed there was some desire to keep the entry names short for other reasons.  By using the elements from the path we get something guaranteed to be unique while keeping some meaning to the name.

serranom:
@chtyim I assume you mean create those jars at runtime?  This is what I had difficulty doing within the test, I think because the classes were getting loaded incorrectly.  I will give that another go.

twill-core/src/main/java/org/apache/twill/internal/ApplicationBundler.java
```diff
@@ -51,6 +34,24 @@
 import java.util.zip.CRC32;
 import java.util.zip.CheckedOutputStream;
 
+import org.apache.twill.api.ClassAcceptor;
```
chtyim: Can you not re-arranging the imports?

twill-core/src/test/java/org/apache/twill/internal/utils/ApplicationBundlerTest.java
```diff
@@ -17,27 +17,32 @@
...
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.jar.JarEntry;
 import java.util.jar.JarInputStream;
+import java.util.jar.JarOutputStream;
+
+import org.apache.commons.compress.utils.IOUtils;
```
chtyim: Same here. The profile that we use have all JDK classes come after other imports.

chtyim:
Other than the import arrangement, the code changes LGTM.

twill-core/src/main/java/org/apache/twill/internal/ApplicationBundler.java
hsaputra: Shouldn't this trigger the style check error?

hsaputra: Ah yes, all tests fail due to checkstyle

serranom:
Fixed these and some other checkstyle related parts.   I was running full builds but was not seeing the checkstyle failures.  I'm not sure what mistake I made.  Anyway, these issues are now fixed.  Thanks for your patience as I get to know the project.  Also, twill-core has these checkstyle warnings but a failure is not triggered:

    [INFO] --- maven-checkstyle-plugin:2.12.1:check (validate) @ twill-core ---
    [INFO] Starting audit...
    /home/martin/dev/other/twill/twill-    core/src/main/java/../java/org/apache/twill/internal/state/SystemMessages.java:22: warning: Wrong order for 'com.google.common.base.Preconditions' import.
    /home/martin/dev/other/twill/twill-core/src/test/java/org/apache/twill/internal/DebugOptionsTest.java:26: warning: First sentence should end with a period.
    Audit done.

twill-core/src/test/java/org/apache/twill/internal/utils/ApplicationBundlerTest.java
```diff
@@ -68,6 +71,56 @@ public void testFindDependencies() throws IOException, ClassNotFoundException {
...
+    Assert.assertSame(classLoader, c1.getClassLoader());
+    Assert.assertSame(classLoader, c2.getClassLoader());
+  }
+
+  private void createJar(Class<?> clazz, File jarFile) throws IOException {
+    try (JarOutputStream jos = new JarOutputStream(new FileOutputStream(jarFile))) {
+      String pathname = clazz.getName().replace(".", "/") + ".class";
+      JarEntry entry = new JarEntry(pathname);
+      jos.putNextEntry(entry);
+      IOUtils.copy(clazz.getClassLoader().getResourceAsStream(pathname), jos);
```
chtyim: Sorry I missed this in last review. Usually we use guava in twill instead of apache common. For this, can you use `Resources.copy(clazz.getClassLoader().getResource(pathname), jos);` instead? Thanks.

chtyim: And that also make sure the proper close of the resource stream.

serranom: I will switch to guava.  And fix the resource stream.

PR #2, opened 2016-08-17
TWILL-190 Wait for Twill runnables to stop when restarting them

poornachandra:
Also, kill non-responding Twill runnables after a timeout.

JIRA - https://issues.apache.org/jira/browse/TWILL-190

Note: This PR is opened against 0.6.0 branch for preliminary review. I'll open another on against master later.

### DIFF
--- a/twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
@@ -22,6 +22,7 @@ import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import org.apache.twill.api.RunId;
@@ -192,13 +193,9 @@ public final class TwillContainerLauncher {
       // TODO: Use configurable value for stop time
       int maxWaitSecs = Constants.APPLICATION_MAX_STOP_SECONDS;
       try {
-        if (shutdownLatch.await(maxWaitSecs, TimeUnit.SECONDS)) {
+        if (Uninterruptibles.awaitUninterruptibly(shutdownLatch, maxWaitSecs, TimeUnit.SECONDS)) {
           return;
         }
-      } catch (InterruptedException e) {
-        LOG.error("Got exception while waiting for runnable {}, instance {} to stop", runnable, instanceId);
-        // TODO: how do we handle the InterruptedException? Should we restore the interrupted status?
-        return;
       } catch (Exception e) {
         LOG.error("Got exception while trying to stop runnable {}, instance {}", runnable, instanceId, e);
       }
@@ -259,33 +256,29 @@ public final class TwillContainerLauncher {
     }
 
     private void killAndWait(int maxWaitSecs) {
-      try {
-        Stopwatch watch = new Stopwatch();
-        watch.start();
-        int tries = 0;
-        while (watch.elapsedTime(TimeUnit.SECONDS) < maxWaitSecs) {
-          // Kill the application
-          try {
-            ++tries;
-            kill();
-          } catch (Exception e) {
-            LOG.error("Exception while killing runnable {}, instance {}", runnable, instanceId, e);
-          }
-
-          // Wait on the shutdownLatch,
-          // if the runnable has stopped then the latch will be count down by completed() method
-          if (shutdownLatch.await(10, TimeUnit.SECONDS)) {
-            // Runnable has stopped now
-            return;
-          }
+      Stopwatch watch = new Stopwatch();
+      watch.start();
+      int tries = 0;
+      while (watch.elapsedTime(TimeUnit.SECONDS) < maxWaitSecs) {
+        // Kill the application
+        try {
+          ++tries;
+          kill();
+        } catch (Exception e) {
+          LOG.error("Exception while killing runnable {}, instance {}", runnable, instanceId, e);
         }
 
-        // Timeout reached, runnable has not stopped
-        LOG.error("Failed to kill runnable {}, instance {} after {} tries", runnable, instanceId, tries);
-        // TODO: should we throw exception here?
-      } catch (InterruptedException e) {
-        LOG.error("Got exception while killing runnable {}, instance {}", runnable, instanceId, e);
+        // Wait on the shutdownLatch,
+        // if the runnable has stopped then the latch will be count down by completed() method
+        if (Uninterruptibles.awaitUninterruptibly(shutdownLatch, 10, TimeUnit.SECONDS)) {
+          // Runnable has stopped now
+          return;
+        }
       }
+
+      // Timeout reached, runnable has not stopped
+      LOG.error("Failed to kill runnable {}, instance {} after {} tries", runnable, instanceId, tries);
+      // TODO: should we throw exception here?
     }
   }
 }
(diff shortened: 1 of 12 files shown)


twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -222,14 +231,17 @@ private void removeInstanceById(String runnableName, int instanceId) {
...
 
+  // This method only stops a runnable using the controller.
+  // The cleanup of the state happens when handleCompleted() method runs for the runnable after the stop
+  // This method will block until handleCompleted() method runs or a timeout occurs
+  // Hence this method should not be called with a containerLock taken
+  private void stopInstanceAndWait(String runnableName, TwillContainerController controller) {
     LOG.info("Stopping service: {} {}", runnableName, controller.getRunId());
+    // This call will block until handleCompleted() method runs or a timeout occurs
     controller.stopAndWait();
-    containers.remove(runnableName, containerId);
```
chtyim: Where do we do the removal of containers now?

poornachandra: It is removed in `handleCompleted()` method.

poornachandra: Realized that we need to clean up state in case of timeout, so added it back.

poornachandra:
@chtyim All the tests pass now, please review.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
```diff
@@ -153,21 +155,29 @@ public TwillContainerController start(RunId runId, int instanceId, Class<?> main
...
     private volatile ContainerLiveNodeData liveData;
 
-    protected TwillContainerControllerImpl(ZKClient zkClient, RunId runId,
+    protected TwillContainerControllerImpl(ZKClient zkClient, RunId runId, String runnable, int instanceId,
                                            ProcessController<Void> processController) {
       super(runId, zkClient);
+      this.runnable = runnable;
+      this.instanceId = instanceId;
       this.processController = processController;
+      this.shutdownLatch = new CountDownLatch(1);
```
hsaputra: I am confused, so the `CountDownLatch` is own as private member and being set in another method within a same instance?

poornachandra: The thread that stops the container needs to wait until the container stops. The only good way to know whether the container has stopped or not is to get a notification from RM during the heartbeat (that happens in a separate thread). Hence we wait on a `CountDownLatch` until the heartbeat thread calls `completed()` method where the latch get count down.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
```diff
@@ -177,7 +187,22 @@ protected void doStartUp() {
...
+        }
+      } catch (InterruptedException e) {
+        LOG.error("Got exception while waiting for runnable {}, instance {} to stop", runnable, instanceId);
+        // TODO: how do we handle the InterruptedException? Should we restore the interrupted status?
+        return;
+      }
+      // Container has not shutdown even after maxWaitSecs after sending stop message,
+      // we'll need to kill the container
+      LOG.warn("Killing runnable {}, instance {} after waiting {} secs", runnable, instanceId, maxWaitSecs);
+      kill();
```
poornachandra: In case of exception during a kill, I'll add code to retry the kill a few times.

hsaputra:
Hi @poornachandra , are you still working on the PR for updates, or ready for review follow up?

poornachandra:
@hsaputra The PR is ready for review. I would appreciate a review of the overall approach and structure. I'm just adding code to better handle exceptions or adding more test cases now. These commits will be small, and will not change the basic structure of the code.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
```diff
@@ -177,7 +188,24 @@ protected void doStartUp() {
...
-      // No-op
+      // Wait for sometime for the container to stop
+      // TODO: Use configurable value for stop time
+      int maxWaitSecs = Constants.APPLICATION_MAX_STOP_SECONDS;
+      try {
+        if (shutdownLatch.await(maxWaitSecs, TimeUnit.SECONDS)) {
+          return;
+        }
+      } catch (InterruptedException e) {
+        LOG.error("Got exception while waiting for runnable {}, instance {} to stop", runnable, instanceId);
```
chtyim: I think it's better to ignore the interrupted exception, or at least clear the thread interrupt flag before calling `await`. This is because

1. you are not blocking forever anyway
2. we do want to spend max effort to let the contain finish it's job before it gets killed.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
```diff
@@ -225,5 +257,35 @@ public ContainerLiveNodeData getLiveNodeData() {
     public void kill() {
       processController.cancel();
     }
+
+    private void killAndWait(int maxWaitSecs) {
+      try {
+        Stopwatch watch = new Stopwatch();
+        watch.start();
+        int tries = 0;
+        while (watch.elapsedTime(TimeUnit.SECONDS) < maxWaitSecs) {
```
chtyim: Why we need to kill it in a loop? Is that mean the `processController.cancel()` doesn't guarantee it will kill the container?

chtyim: If we have to loop, it's better to have it inside the `ProcessController.cancel()` method.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -256,20 +261,36 @@ public void acquired(List<? extends ProcessLauncher<YarnContainerInfo>> launcher
...
+    ExecutorService stopPoller = Executors.newSingleThreadExecutor(Threads.createDaemonThreadFactory("stopPoller"));
+    stopPoller.execute(new Runnable() {
+      @Override
+      public void run() {
+        while (!ids.isEmpty()) {
+          try {
+            amClient.allocate(0.0f, handler);
+            TimeUnit.SECONDS.sleep(1);
+          } catch (InterruptedException e) {
+            // Exit on an interrupt
```
chtyim: This comment is not accurate.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -846,24 +875,31 @@ private boolean handleRestartRunnablesInstances(final Message message, final Run
   /**
    * Helper method to restart instances of runnables.
    */
-  private void restartRunnableInstances(String runnableName, @Nullable Set<Integer> instanceIds) {
+  private void restartRunnableInstances(String runnableName, @Nullable Set<Integer> instanceIds)
+    throws InterruptedException {
```
chtyim: I don't think this method throws `InterruptedException`. This also mean we don't have to change the `handleRestartRunnablesInstances` method.

poornachandra: Good point, looks like I missed it during a refactor. Will remove it.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
poornachandra: Done

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
poornachandra: Agreed, I'll make the change

poornachandra:
@chtyim Except for moving the kill loop to `ProcessController.cancel()` method, I have addressed all other comments.

PR #4, opened 2016-08-25
TWILL-190 Wait for Twill runnables to stop when restarting them

poornachandra:
Also kill non-responding Twill runnables after a timeout.

JIRA - https://issues.apache.org/jira/browse/TWILL-190

Opening PR https://github.com/apache/twill/pull/2 against master this time

### DIFF
--- a/twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
@@ -18,9 +18,11 @@
 package org.apache.twill.internal;
 
 import com.google.common.base.Charsets;
+import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.gson.Gson;
 import com.google.gson.JsonElement;
 import org.apache.twill.api.RunId;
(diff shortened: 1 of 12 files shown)


twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
```diff
@@ -220,5 +250,31 @@ public ContainerLiveNodeData getLiveNodeData() {
...
+        // Wait on the shutdownLatch,
+        // if the runnable has stopped then the latch will be count down by completed() method
+        if (Uninterruptibles.awaitUninterruptibly(shutdownLatch, 10, TimeUnit.SECONDS)) {
+          // Runnable has stopped now
+          return;
+        }
+      }
+
+      // Timeout reached, runnable has not stopped
+      LOG.error("Failed to kill runnable {}, instance {} after {} tries", runnable, instanceId, tries);
```
chtyim: Showing the number of tries is quite artificial since the retry is based on time. I think it's better to just say failed to kill after n seconds.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -268,20 +270,33 @@ public void acquired(List<? extends ProcessLauncher<YarnContainerInfo>> launcher
...
+    // Handle heartbeats during shutdown because runningContainers.stopAll() waits until
+    // handleCompleted() is called for every stopped runnable
+    ExecutorService stopPoller = Executors.newSingleThreadExecutor(Threads.createDaemonThreadFactory("stopPoller"));
+    stopPoller.execute(new Runnable() {
+      @Override
+      public void run() {
+        while (!ids.isEmpty()) {
+          try {
+            amClient.allocate(0.0f, handler);
+            TimeUnit.SECONDS.sleep(1);
```
chtyim: Should check if `ids` is already emptied before sleeping, since the call the `allocate` may already have the ids emptied by the handler and we don't have the sleep for an extra second for that.

chtyim:
LGTM

PR #6, opened 2016-08-26
TWILL-175 Avoid caching YarnClient

anwar6953:
Avoid caching a YarnClient, since each YarnClient is fixed to a particular UserGroupInformation.

Note: The diff is very similar as to the changes made to the same file in:
https://github.com/caskdata/cdap/pull/6024/files#diff-41467cacd78ce3eba37c5aa839c4a508
Additional change had to be made to the `addRMToken` method, since that doesn't exist in the CDAP repo.

https://issues.apache.org/jira/browse/TWILL-175

### DIFF
--- a/twill-yarn/src/main/hadoop20/org/apache/twill/internal/yarn/Hadoop20YarnAppClient.java
+++ b/twill-yarn/src/main/hadoop20/org/apache/twill/internal/yarn/Hadoop20YarnAppClient.java
@@ -59,21 +59,29 @@ import javax.annotation.Nullable;
  * Apache Hadoop 2.0.
  * </p>
  */
-public final class Hadoop20YarnAppClient extends AbstractIdleService implements YarnAppClient {
+public final class Hadoop20YarnAppClient implements YarnAppClient {
 
   private static final Logger LOG = LoggerFactory.getLogger(Hadoop20YarnAppClient.class);
-  private final YarnClient yarnClient;
+  private final Configuration configuration;
   private String user;
 
   public Hadoop20YarnAppClient(Configuration configuration) {
-    this.yarnClient = new YarnClientImpl();
-    yarnClient.init(configuration);
+    this.configuration = configuration;
     this.user = System.getProperty("user.name");
+  }
+
+  // Creates and starts a yarn client
+  private YarnClient createYarnClient() {
+    YarnClient yarnClient = new YarnClientImpl();
+    yarnClient.init(configuration);
+    yarnClient.start();
+    return yarnClient;
   }
 
   @Override
   public ProcessLauncher<ApplicationMasterInfo> createLauncher(TwillSpecification twillSpec,
                                                                @Nullable String schedulerQueue) throws Exception {
+    final YarnClient yarnClient = createYarnClient();
     // Request for new application
     final GetNewApplicationResponse response = yarnClient.getNewApplication();
     final ApplicationId appId = response.getApplicationId();
@@ -100,7 +108,7 @@ public final class Hadoop20YarnAppClient extends AbstractIdleService implements 
       @Override
       public ProcessController<YarnApplicationReport> submit(YarnLaunchContext launchContext) {
         ContainerLaunchContext context = launchContext.getLaunchContext();
-        addRMToken(context);
+        addRMToken(context, yarnClient);
         context.setUser(appSubmissionContext.getUser());
         context.setResource(adjustMemory(response, capability));
         appSubmissionContext.setAMContainerSpec(context);
@@ -131,7 +139,7 @@ public final class Hadoop20YarnAppClient extends AbstractIdleService implements 
     return capability;
   }
 
-  private void addRMToken(ContainerLaunchContext context) {
+  private void addRMToken(ContainerLaunchContext context, YarnClient yarnClient) {
     if (!UserGroupInformation.isSecurityEnabled()) {
       return;
     }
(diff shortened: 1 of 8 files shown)


chtyim:
You need to make similar change to the `Hadoop20YarnAppClient`. You can enable it by enabling the hadoop-2.0 profile (and disable other hadoop-* profiles).

twill-yarn/src/main/hadoop21/org/apache/twill/internal/yarn/Hadoop21YarnAppClient.java
```diff
@@ -127,6 +135,7 @@ private void addRMToken(ContainerLaunchContext context) {
       return;
     }
 
+    YarnClient yarnClient = createYarnClient();
```
chtyim: This `addRMToken` method only gets called from the `ApplicationSubmitter.submit()` method, which only get triggered form the `createLauncher` method. Since there is already a new `YarnClient` created inside the `createLauncher` method, you can just have that one passed to this `addRMToken` method and use it instead of creating another one here.

twill-yarn/src/main/hadoop21/org/apache/twill/internal/yarn/Hadoop21YarnAppClient.java
```diff
@@ -58,16 +58,24 @@
...
+    YarnClient yarnClient = YarnClient.createYarnClient();
     yarnClient.init(configuration);
+    yarnClient.start();
+    return yarnClient;
   }
 
   @Override
   public ProcessLauncher<ApplicationMasterInfo> createLauncher(TwillSpecification twillSpec,
                                                                @Nullable String schedulerQueue) throws Exception {
+    final YarnClient yarnClient = createYarnClient();
```
chtyim: When will this get stopped?

twill-yarn/src/main/hadoop21/org/apache/twill/internal/yarn/Hadoop21YarnAppClient.java
```diff
@@ -197,6 +213,7 @@ public YarnApplicationReport getReport() {
     public void cancel() {
       try {
         yarnClient.killApplication(appId);
+        yarnClient.stop();
```
chtyim: This method will not get called if the container stopped gracefully.

twill-yarn/src/main/hadoop21/org/apache/twill/internal/yarn/Hadoop21YarnAppClient.java
anwar6953: It gets passed into and used by the `ProcessControllerImpl` at the end of the `createLauncher` method.
It gets stopped when `cancel` is called on the `ProcessControllerImpl` (defined at the bottom of this class).

twill-yarn/src/main/hadoop21/org/apache/twill/internal/yarn/Hadoop21YarnAppClient.java
anwar6953: I see. Any hook to close the YarnClient, otherwise? This is what I found.

anwar6953: Alternatively, ProcessControllerImpl#getReport can create+start and stop it each time it gets a report.

chtyim: You have to keep this `yarnClient` for killing though since it is binding to the user who launched the app.

You probably have to make `ProcessController` `AutoCloseable` so that on application completion, you can close it from the `YarnTwillController`.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunnableProcessLauncher.java
```diff
@@ -76,6 +76,11 @@ public String toString() {
 
     return new ProcessController<R>() {
       @Override
+      public void close() throws Exception {
+        // no-op
```
anwar6953: is it fine to leave the runnable process running upon Close of the controller?

chtyim: Controller ties to the lifecycle of the stuff that it is controlling. What do mean by leave the runnable process running when closing the controller? Is it related to #4 ?

anwar6953: I just meant, now both close and cancel will kill the runnable?

chtyim: No. Close should just release resources

anwar6953: So then the close method will be a no-op in this case.

twill-yarn/src/main/hadoop20/org/apache/twill/internal/yarn/Hadoop20YarnAppClient.java
```diff
@@ -62,18 +62,26 @@
...
+    YarnClient yarnClient = new YarnClientImpl();
+    yarnClient.init(configuration);
+    yarnClient.start();
+    return yarnClient;
+  }
+
   @Override
   public ProcessLauncher<ApplicationMasterInfo> createLauncher(TwillSpecification twillSpec,
                                                                @Nullable String schedulerQueue) throws Exception {
+    final YarnClient yarnClient = createYarnClient();
```
chtyim: Why this one will get closed?

chtyim: Nevermind, I found it.

twill-yarn/src/main/hadoop20/org/apache/twill/internal/yarn/Hadoop20YarnAppClient.java
```diff
@@ -176,22 +184,27 @@ private void addRMToken(ContainerLaunchContext context) {
...
+      return yarnClient.getNodeReports();
+    } finally {
+      yarnClient.stop();
+    }
   }
 
   @Override
   protected void startUp() throws Exception {
-    yarnClient.start();
+    // no-op. We will create and start YarnClients, on demand
```
chtyim: Maybe we should make `YarnAppClient` not a `Service` at all. It seems like there is no need for any lifecycle management.

chtyim:
LGTM. Can you squash the commits?

PR #7, opened 2016-08-26
TWILL-185 Allow user to avoid scheduling the LocationSecureStoreUpdater.

anwar6953:
This will allow the use case of user scheduling their own secure store updates.
As a workaround, for example:
https://github.com/caskdata/cdap/pull/6380/commits/e77b9ac1fa6c9176a49cb27d6118e2d3065b93cd#diff-832334ca819881b7884d50946c76ce00L332

It will also alleviate some of the issues described in https://issues.apache.org/jira/browse/TWILL-109.

https://issues.apache.org/jira/browse/TWILL-185

### DIFF
--- a/twill-core/src/main/java/org/apache/twill/internal/Configs.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/Configs.java
@@ -31,6 +31,11 @@ public final class Configs {
      */
     public static final String JAVA_RESERVED_MEMORY_MB = "twill.java.reserved.memory.mb";
 
+    /**
+     * Set this to false to disable the secure store updates done by default.
+     */
+    public static final String SECURE_STORE_UPDATE_LOCATION_ENABLED = "twill.secure.store.update.location.enabled";
+
     private Keys() {
     }
   }
--- a/twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillRunnerService.java
+++ b/twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillRunnerService.java
@@ -65,6 +65,7 @@ import org.apache.twill.common.Threads;
 import org.apache.twill.filesystem.FileContextLocationFactory;
 import org.apache.twill.filesystem.Location;
 import org.apache.twill.filesystem.LocationFactory;
+import org.apache.twill.internal.Configs;
 import org.apache.twill.internal.Constants;
 import org.apache.twill.internal.ProcessController;
 import org.apache.twill.internal.RunIds;
@@ -330,8 +331,9 @@ public final class YarnTwillRunnerService implements TwillRunnerService {
     watchCancellable = watchLiveApps();
     liveInfos = createLiveInfos();
 
+    boolean enableSecureStoreUpdate = yarnConfig.getBoolean(Configs.Keys.SECURE_STORE_UPDATE_LOCATION_ENABLED, true);
     // Schedule an updater for updating HDFS delegation tokens
-    if (UserGroupInformation.isSecurityEnabled()) {
+    if (UserGroupInformation.isSecurityEnabled() && enableSecureStoreUpdate) {
       long renewalInterval = yarnConfig.getLong(DFSConfigKeys.DFS_NAMENODE_DELEGATION_TOKEN_RENEW_INTERVAL_KEY,
                                                 DFSConfigKeys.DFS_NAMENODE_DELEGATION_TOKEN_RENEW_INTERVAL_DEFAULT);
       // Schedule it five minutes before it expires.


twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillRunnerService.java
```diff
@@ -211,7 +226,7 @@ public String getJVMOptions() {
   @Override
   public Cancellable scheduleSecureStoreUpdate(final SecureStoreUpdater updater,
                                                long initialDelay, long delay, TimeUnit unit) {
-    if (!UserGroupInformation.isSecurityEnabled()) {
+    if (!UserGroupInformation.isSecurityEnabled() || !enableSecureStoreUpdate) {
```
chtyim: You shouldn't disable this one. Otherwise user won't be able to schedule his own update. We should only disable the default one by Twill (the one that update HDFS delegation tokens in `startUp`).

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillRunnerService.java
```diff
@@ -157,11 +158,25 @@ public YarnTwillRunnerService(YarnConfiguration config, String zkConnect) {
...
+   * @param enableSecureStoreUpdate Whether to update SecureStore periodically
+   */
+  public YarnTwillRunnerService(YarnConfiguration config, String zkConnect, LocationFactory locationFactory,
+                                boolean enableSecureStoreUpdate) {
     this.yarnConfig = config;
     this.yarnAppClient = new VersionDetectYarnAppClientFactory().create(config);
     this.locationFactory = locationFactory;
     this.zkClientService = getZKClientService(zkConnect);
     this.controllers = HashBasedTable.create();
+    this.enableSecureStoreUpdate = enableSecureStoreUpdate;
```
chtyim: Better use a configuration passed in from the `YarnConfiguration`. The keys should be defined in `Configs.Keys` class.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillRunnerService.java
anwar6953: oh right; good point.

twill-core/src/main/java/org/apache/twill/internal/Configs.java
```diff
@@ -31,6 +31,11 @@
      */
     public static final String JAVA_RESERVED_MEMORY_MB = "twill.java.reserved.memory.mb";
 
+    /**
+     *
+     */
+    public static final String ENABLE_DEFAULT_SECURE_STORE_UPDATE = "twill.enable.default.secure.store.update";
```
hsaputra: Could you add description what this constant for?

twill-core/src/main/java/org/apache/twill/internal/Configs.java
```diff
@@ -31,6 +31,11 @@
      */
     public static final String JAVA_RESERVED_MEMORY_MB = "twill.java.reserved.memory.mb";
 
+    /**
+     * Set this to false to disable the secure store updates done by default.
+     */
+    public static final String ENABLE_DEFAULT_SECURE_STORE_UPDATE = "twill.enable.default.secure.store.update";
```
chtyim: I think the common convention is to call it like `twill.secure.store.update.location.enabled` (from widest to narrowest scope, see hadoop-default.xml). Explicitly calling out `location` because that is what get updated by default right now and potentially there can be other secure store gets updated by default.

hsaputra: +1

anwar6953:
squashed commits

twill-core/src/main/java/org/apache/twill/internal/Configs.java
```diff
@@ -31,6 +31,11 @@
      */
     public static final String JAVA_RESERVED_MEMORY_MB = "twill.java.reserved.memory.mb";
 
+    /**
+     * Set this to false to disable the secure store updates done by default.
+     */
+    public static final String ENABLE_DEFAULT_SECURE_STORE_UPDATE = "twill.secure.store.update.location.enabled";
```
chtyim: It's better to have the constant name matches with the value. E.g. `SECURE_STORE_UPDATE_LOCATION_ENABLED`

anwar6953: Updated and squashed.
This was the only diff:
image

chtyim:
LGTM.

PR #9, opened 2016-08-31
(TWILL-173) Have EmbeddedKafkaServer restart multiple times on bind failure

chtyim:
- Due to potential race condition between random port generation vs actual binding, there is a possibility that the binding would fail.

### DIFF
--- a/twill-core/src/main/java/org/apache/twill/internal/kafka/EmbeddedKafkaServer.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/kafka/EmbeddedKafkaServer.java
@@ -59,7 +59,7 @@ public final class EmbeddedKafkaServer extends AbstractIdleService {
   protected void startUp() throws Exception {
     int tries = 0;
     do {
-      KafkaConfig kafkaConfig = new KafkaConfig(properties);
+      KafkaConfig kafkaConfig = createKafkaConfig(properties);
       KafkaServer kafkaServer = createKafkaServer(kafkaConfig);
       try {
         kafkaServer.startup();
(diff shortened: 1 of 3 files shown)


twill-core/src/main/java/org/apache/twill/internal/kafka/EmbeddedKafkaServer.java
```diff
@@ -65,9 +72,19 @@ protected void startUp() throws Exception {
...
           throw e;
         }
+
+        // Do a random sleep of < 200ms
+        TimeUnit.MILLISECONDS.sleep(new Random().nextInt(200) + 1L);
+
+        // Generate a new port for the Kafka
+        int port = Networks.getRandomPort();
+        Preconditions.checkState(port > 0, "Failed to get random port.");
+        properties.setProperty("port", Integer.toString(port));
```
albertshau: so "port" is required to be in the Properties passed to KafkaConfig? Is there any place where a port is being set in the Properties passed to the constructor, and then used somewhere else?

chtyim: Currently it's set in the ApplicationMasterService. Probably better to have it left empty (or set to 0) in order to trigger this logic.

albertshau: Should we only do this if its originally set to 0 or left empty? Without reading the code I would expect it to connect to the port I set, or not connect at all.

Or is this not a concern because the port is not used to connect?

albertshau:
lgtm

PR #11, opened 2016-09-21
Fix a ClassCastException in YarnUtils.addDelegationTokens.

anwar6953:
Earlier in the method, we `unwrap` the `locationFactory`. Also, the `instanceof` check is on the unwrapped object. So, we should cast the unwrapped object, not the original object.

Example exception:
```
java.lang.ClassCastException: org.apache.twill.filesystem.LocationFactories$1 cannot be cast to org.apache.twill.filesystem.FileContextLocationFactory
        at org.apache.twill.internal.yarn.YarnUtils.addDelegationTokens(YarnUtils.java:168) ~[org.apache.twill.twill-yarn-0.8.0-SNAPSHOT.jar:0.8.0-SNAPSHOT]
```

### DIFF
--- a/twill-yarn/src/main/java/org/apache/twill/internal/yarn/YarnUtils.java
+++ b/twill-yarn/src/main/java/org/apache/twill/internal/yarn/YarnUtils.java
@@ -165,7 +165,7 @@ public class YarnUtils {
         tokens = ImmutableList.copyOf(fsTokens);
       }
     } else if (factory instanceof FileContextLocationFactory) {
-      FileContext fc = ((FileContextLocationFactory) locationFactory).getFileContext();
+      FileContext fc = ((FileContextLocationFactory) factory).getFileContext();
       tokens = fc.getDelegationTokens(new Path(locationFactory.create("/").toURI()), renewer);
     }
 


twill-yarn/src/main/java/org/apache/twill/internal/yarn/YarnUtils.java
```diff
@@ -165,7 +165,7 @@ public static ApplicationId createApplicationId(long timestamp, int id) {
         tokens = ImmutableList.copyOf(fsTokens);
       }
     } else if (factory instanceof FileContextLocationFactory) {
-      FileContext fc = ((FileContextLocationFactory) locationFactory).getFileContext();
+      FileContext fc = ((FileContextLocationFactory) factory).getFileContext();
```
bdmogal: is there no test that can trigger this failure?

bdmogal:
fix lgtm. wondering if this code path is not executed in tests though.

twill-yarn/src/main/java/org/apache/twill/internal/yarn/YarnUtils.java
anwar6953: There's no end-to-end unit test cases in the twill-yarn module. And there's no unit tests for this class.
Any integration test on a secure cluster would've triggered this failure though.
I've verified that the original issue is no longer hit after I made this change.

anwar6953:
I think the failed test cases (https://travis-ci.org/apache/twill/builds/161527818) are flaky.
Rerunning to check.

poornachandra:
LGTM

PR #14, opened 2016-10-05
TWILL-138 Runtime log level change

yaojiefeng:
JIRA: https://issues.apache.org/jira/browse/TWILL-138

Summary: Make API changes to TwillController and TwillPreparer to set the log level for the twill application or specific twill runnables at runtime or before the application starts. The TwillPreparer will serialize the log level change requirements and deserialize them once service starts. The TwillController will use messages to send out log level change request. Details for implementation can be found in JIRAs.


twill-api/src/main/java/org/apache/twill/api/ResourceReport.java
```diff
@@ -61,4 +63,19 @@
    * @return list of services of the application master.
    */
   List<String> getServices();
+
+  /**
+   * Get the map of the log level arguments of the twill application.
+   *
+   * @return the map of the log level arguments of the twill application
+   */
+  Map<String, Map<String, LogEntry.Level>> getLogLevelArguments();
```
chtyim: What's the contract? What's the key? Does it always return all log levels even nothing was set?

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,29 @@
    * @return A {@link Future} that will be completed when the restart operation has been done.
    */
   Future<String> restartInstances(String runnable, int instanceId, int... moreInstanceIds);
+
+  /**
+   * Set the root log level for Twill applications running in a container.
```
chtyim: In a container? Which container? Isn't that this affect all containers of all runnables as well as the AM?

twill-api/src/main/java/org/apache/twill/api/TwillPreparer.java
```diff
@@ -227,7 +227,7 @@
   TwillPreparer addSecureStore(SecureStore secureStore);
 
   /**
-   * Set the log level for Twill applications running in a container.
+   * Set the root log level for Twill applications running in a container.
```
chtyim: Same here

twill-api/src/main/java/org/apache/twill/api/TwillRunResources.java
```diff
@@ -59,6 +61,16 @@
   /**
    * @return the enabled log level for the container where the runnable is running in.
    */
-  LogEntry.Level getLogLevel();
+  LogEntry.Level getRootLogLevel();
+
+  /**
+   * @return the enabled log level arguments for the container where the runnable is running in.
+   */
+  Map<String, LogEntry.Level> getLogLevelArguments();
```
chtyim: What do mean by "LogLevelArguments"? Should be "getLogLevels".

twill-api/src/main/java/org/apache/twill/api/TwillRunResources.java
```diff
@@ -59,6 +61,16 @@
...
 
+  /**
+   * Update the log level arguments for the container where the runnable is running in.
+   * @param logLevelArguments
+   */
+  void updateLogLevel(Map<String, LogEntry.Level> logLevelArguments);
```
chtyim: Why there is an update method? This interface is supposedly for reporting, hence is immutable.

twill-api/src/main/java/org/apache/twill/internal/DefaultTwillRunResources.java
```diff
@@ -30,17 +35,26 @@
...
+                                  Level rootLogLevel, Map<String, LogEntry.Level> logLevelArguments) {
     this.instanceId = instanceId;
     this.containerId = containerId;
     this.virtualCores = cores;
     this.memoryMB = memoryMB;
     this.host = host;
     this.debugPort = debugPort;
-    this.logLevel = logLevel;
+    this.rootLogLevel = rootLogLevel;
+    this.logLevelArguments = logLevelArguments;
```
chtyim: should just call `logLevels`.

twill-api/src/main/java/org/apache/twill/api/TwillPreparer.java
```diff
@@ -236,6 +236,23 @@
   TwillPreparer setLogLevel(LogEntry.Level logLevel);
 
   /**
+   * Set the log levels for requested logger names for Twill applications running in a container.
+   *
+   * @param logLevelAppArgs The {@link Map} contains the requested logger names and log levels that need to be set.
+   * @return This {@link TwillPreparer}.
+   */
+  TwillPreparer setLogLevel(Map<String, LogEntry.Level> logLevelAppArgs);
```
chtyim: `setLogLevels`.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,29 @@
...
+  Future<String> setLogLevel(Map<String, LogEntry.Level> logLevelAppArgs);
+
+  /**
+   * Set the log levels for requested logger names for a {@link TwillRunnable}.
+   *
+   * @param runnableName The name of the runnable to set the log level.
+   * @param logLevels The {@link Map} contains the requested logger name and log level that need to be changed.
+   * @return A {@link Future} that will be completed when the set log level operation has been done.
+   */
+  Future<String> setLogLevel(String runnableName, Map<String, LogEntry.Level> logLevels);
```
chtyim: `setLogLevels`.

twill-api/src/main/java/org/apache/twill/internal/DefaultTwillRunResources.java
```diff
@@ -90,8 +104,21 @@ public Integer getDebugPort() {
...
+    return rootLogLevel;
+  }
+
+  @Override
+  public Map<String, LogEntry.Level> getLogLevelArguments() {
+    return logLevelArguments;
+  }
+
+  @Override
+  public void updateLogLevel(Map<String, Level> logLevelArguments) {
```
chtyim: Better keep this class immutable. If there is a change in the log level, simply create a new instance.

twill-common/src/main/java/org/apache/twill/internal/Constants.java
```diff
@@ -82,6 +82,14 @@ private Files() {
     }
   }
 
+  /**
+   * Constants for logLevel system messages.
+   */
+  public static final class SystemMessages {
```
chtyim: Why need string constants here? All system messages/commands should be defined in `SystemMessages`. You can use object comparison instead of string comparison.

twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
```diff
@@ -137,7 +138,7 @@ public final synchronized ServiceDiscovered discoverService(String serviceName)
 
   @Override
   public final ListenableFuture<Set<String>> restartInstances(Map<String,
-      ? extends Set<Integer>> runnableToInstanceIds) {
+    ? extends Set<Integer>> runnableToInstanceIds) {
```
chtyim: This new line is awkward, even in the old code. The parent actually returns `Future` instead of returning `ListenableFuture`. If you change the return type to `Future`, everything should fits on one line.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -72,7 +76,7 @@ public String getZkConnectStr() {
     return zkConnectStr;
   }
 
-  public RunId getTwillRunId() {
+  public RunId getTwillAppRunId() {
```
chtyim: There is only one `RunId`, why not just call if `runId` and `getRunId` for the field name and method name respectively?

twill-core/src/main/java/org/apache/twill/internal/state/SystemMessages.java
```diff
@@ -68,6 +73,21 @@ public static Message updateRunnablesInstances(Command updateCommand) {
     return new SimpleMessage(Message.Type.SYSTEM, Message.Scope.RUNNABLES, null, updateCommand);
   }
 
+  public static Message setLogLevel(Map<String, LogEntry.Level> logLevelArguments) {
+    return setLogLevel(Constants.SystemMessages.LOG_ALL_RUNNABLES, logLevelArguments);
```
chtyim: Why not use the `Message.Scope.RUNNABLES`?

twill-core/src/main/java/org/apache/twill/internal/utils/LogLevelUtil.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Collections of helper functions for log level change.
+ */
+public final class LogLevelUtil {
+
+  /**
+   * Convert the log level argument value type to string.
+   */
+  public static Map<String, String> convertLogLevelValuesToString(Map<String, LogEntry.Level> logLevelArguments) {
```
chtyim: Seems like this method only used in `SystemMessages`. Better move it there as a private method. Haven't too many "util" class is not good.

twill-core/src/main/java/org/apache/twill/internal/utils/LogLevelUtil.java
```diff
@@ -0,0 +1,85 @@
...
+      public String transformEntry(String loggerName, LogEntry.Level level) {
+        return level.name();
+      }
+    });
+  }
+
+  /**
+   * Convert the log level argument type to LogEntry.Level
+   */
+  public static Map<String, LogEntry.Level> convertLogLevelValuesToLogEntry(Map<String, String> logLevelArguments) {
```
chtyim: Similarly this one, it's better having a method in `SystemMessages` that extract the log level map from a `Messages`.

twill-core/src/main/java/org/apache/twill/internal/utils/LogLevelUtil.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+   * Get the log level arguments for a twill runnable.
+   *
+   * @param runnableName name of the runnable.
+   * @param logLevelArguments the arguments for all runnables.
+   * @return the map of the log level arguments for the runnable, empty if there is no argument.
+   */
+  public static Map<String, LogEntry.Level> getLogLevelForRunnable(
```
chtyim: Same here.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
yaojiefeng: I changed it because in {code}TwillContainerMain{code}, the main method will try to get the run id for both the application and the run id for the TwillRunnable. And since TwillRuntimeSpecification is mainly to pass runtime information for TwillApplication, I feel like it is clearer to name like this.

twill-api/src/main/java/org/apache/twill/api/ResourceReport.java
yaojiefeng: The key of the map is the runnable name, and the value of the map is a map of logger name and log level pair. If nothing is set for log level, it will return an empty map.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
yaojiefeng: Yes. It will affect all containers, I will change the javadoc

twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
yaojiefeng: I cannot change the type to Future for this method, there is a another method calling it which requires the return type to be a ListenableFuture

twill-core/src/main/java/org/apache/twill/internal/utils/LogLevelUtil.java
yaojiefeng: The methods in `LogLevelUtil` are not only used in `SystemMessages`. They are also used in `RunningContainers` and `TwillContainerMain`, is it good to have all these methods in `SystemMessages`?

twill-core/src/main/java/org/apache/twill/internal/state/SystemMessages.java
yaojiefeng: The constants about log levels are only used for `Messages`, they are also used in `TwillPreparer`.

twill-yarn/src/test/java/org/apache/twill/yarn/LogLevelChangeTestRun.java
```diff
@@ -0,0 +1,237 @@
...
+
+    @Override
+    public void run() {
+      this.runThread = Thread.currentThread();
+
+      // check if the initial log level is DEBUG
+      Assert.assertTrue(LOG.isDebugEnabled() && !LOG.isTraceEnabled());
+
+      int i = 0;
+      while (!Thread.interrupted()) {
```
chtyim: Without sleeping/pausing, the thread can take up all CPU resources, causing starvation for other threads. On machine with single core, unit-test could fail due to this.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -298,6 +299,29 @@ public TwillPreparer addSecureStore(SecureStore secureStore) {
   public TwillPreparer setLogLevel(LogEntry.Level logLevel) {
     Preconditions.checkNotNull(logLevel);
     this.logLevel = logLevel;
+    Map<String, LogEntry.Level> logLevels = new HashMap<>();
+    logLevels.put(Logger.ROOT_LOGGER_NAME, logLevel);
+    saveLogLevels(Constants.SystemMessages.LOG_ALL_RUNNABLES, logLevels);
```
chtyim: Why use the `LOG_ALL_RUNNABLES` as a runnable name? We have the `twillSpec` in this class, so it's easy to just expand the log level settings for each runnable.

twill-yarn/src/main/java/org/apache/twill/internal/container/TwillContainerService.java
```diff
@@ -17,26 +17,41 @@
...
 import org.apache.twill.internal.yarn.AbstractYarnTwillService;
 import org.apache.twill.zookeeper.ZKClient;
+import org.slf4j.ILoggerFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.Reader;
```
chtyim: Unused imports?

twill-core/src/main/java/org/apache/twill/internal/utils/LogLevelUtil.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Get the log level arguments for a twill runnable.
+   *
+   * @param runnableName name of the runnable.
+   * @param logLevels the arguments for all runnables.
+   * @return the map of the log level arguments for the runnable, empty if there is no argument.
+   */
+  public static Map<String, LogEntry.Level> getLogLevelForRunnable(
```
chtyim: We shouldn't have this method as we shouldn't use `LOG_ALL_RUNNABLES` as a special runnable name. It's better to have the log levels populated to each runnable log level map when setting/changing it.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
```diff
@@ -84,22 +84,29 @@ public TwillContainerLauncher(RuntimeSpecification runtimeSpec, ContainerInfo co
    * @param classPath The class path to load classes for the container.
    * @return instance of {@link TwillContainerController} to control the container run.
    */
-  public TwillContainerController start(RunId runId, int instanceId, Class<?> mainClass, String classPath) {
+  public TwillContainerController start(RunId runId, int instanceId, Class<?> mainClass, String classPath,
+                                        Location logLevelLocation) {
```
chtyim: Is `logLevelLocation` nullable? I see you check that in line 104. If that's the case, annotate this parameter with `@Nullable`. Also, please update the javadoc accordingly.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -40,11 +41,13 @@
   private final int reservedMemory;
   private final String rmSchedulerAddr;
   private final LogEntry.Level logLevel;
+  private final Map<String, Map<String, LogEntry.Level>> logLevels;
 
   public TwillRuntimeSpecification(TwillSpecification twillSpecification, String fsUser, URI twillAppDir,
                                    String zkConnectStr, RunId twillRunId, String twillAppName,
                                    int reservedMemory, @Nullable String rmSchedulerAddr,
-                                   @Nullable LogEntry.Level logLevel) {
+                                   @Nullable LogEntry.Level logLevel,
```
chtyim: better rename this as `rootLogLevel`, just like in other classes. Also rename the getter.

twill-core/src/main/java/org/apache/twill/internal/json/TwillRunResourcesCodec.java
```diff
@@ -41,7 +43,8 @@
   private static final String MEMORY_MB = "memoryMB";
   private static final String VIRTUAL_CORES = "virtualCores";
   private static final String DEBUG_PORT = "debugPort";
-  private static final String LOG_LEVEL = "logLevel";
+  private static final String LOG_LEVEL = "rootLogLevel";
+  private static final String LOG_LEVEL_ARGUMENTS = "logLevels";
```
chtyim: Why not call it `LOG_LEVELS`?

twill-core/src/main/java/org/apache/twill/internal/json/TwillRunResourcesCodec.java
```diff
@@ -41,7 +43,8 @@
   private static final String MEMORY_MB = "memoryMB";
   private static final String VIRTUAL_CORES = "virtualCores";
   private static final String DEBUG_PORT = "debugPort";
-  private static final String LOG_LEVEL = "logLevel";
+  private static final String LOG_LEVEL = "rootLogLevel";
```
chtyim: Should rename to `ROOT_LOG_LEVEL` to be consistent.

twill-core/src/main/java/org/apache/twill/internal/json/TwillRunResourcesCodec.java
```diff
@@ -55,9 +58,11 @@ public JsonElement serialize(TwillRunResources src, Type typeOfSrc, JsonSerializ
     if (src.getDebugPort() != null) {
       json.addProperty(DEBUG_PORT, src.getDebugPort());
     }
-    if (src.getLogLevel() != null) {
-      json.addProperty(LOG_LEVEL, src.getLogLevel().toString());
+    if (src.getRootLogLevel() != null) {
+      json.addProperty(LOG_LEVEL, src.getRootLogLevel().toString());
```
chtyim: for serializing enum, always use `.name()` instead of `.toString()`.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,30 @@
    * @return A {@link Future} that will be completed when the restart operation has been done.
    */
   Future<String> restartInstances(String runnable, int instanceId, int... moreInstanceIds);
+
+  /**
+   * Set the root log level for Twill applications in all containers.
+   *
+   * @param logLevel The log level for the root logger to change.
+   * @return A {@link Future} that will be completed when the set log level operation has been done.
```
yaojiefeng: The `String` in `Future` in `setLogLevel` and `setLogLevels` for the application is the string representation of the log levels passed in, the `String` in `Future` for `setLogLevels` for a runnable is the runnableName.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -200,7 +201,9 @@ private RunningContainers initRunningContainers(ContainerId appMasterContainerId
       Integer.parseInt(System.getenv(EnvKeys.YARN_CONTAINER_MEMORY_MB)),
       appMasterHost, null, null);
     String appId = appMasterContainerId.getApplicationAttemptId().getApplicationId().toString();
-    return new RunningContainers(appId, appMasterResources, zkClient);
+    return new RunningContainers(appId, appMasterResources, zkClient,
+                                 twillRuntimeSpec.getLogLevels(), applicationLocation,
+                                 new ArrayList<>(twillSpec.getRunnables().keySet()));
```
chtyim: No need to copy since twill spec is immutable.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -951,4 +958,34 @@ public void run() {
       }
     };
   }
+
+  /**
+   * Attempt to change the log level from a runnable or all runnables.
+   *
+   * @return {@code true} if the message requests changing log levels and {@code false} otherwise.
+   */
+  private boolean handleLogLevelChange(final Message message, final Runnable completion) {
```
chtyim: Seems like the `final` is unnecessary.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -951,4 +958,34 @@ public void run() {
...
+    Command command = message.getCommand();
+    if (!command.getCommand().equals(SystemMessages.LOG_LEVEL)) {
+      return false;
+    }
+
+    if (scope == Message.Scope.ALL_RUNNABLE) {
+      runningContainers.sendToAll(message, completion);
+    } else {
+      final String runnableName = message.getRunnableName();
+      if (runnableName == null || runnableName.isEmpty() || !twillSpec.getRunnables().containsKey(runnableName)) {
```
chtyim: No need to check if `runnableName` is empty.

twill-core/src/main/java/org/apache/twill/internal/DefaultResourceReport.java
```diff
@@ -149,6 +161,51 @@ public void setServices(List<String> services) {
...
+  }
+
+  /**
+   * Save the log levels for the twill runnable.
+   *
+   * @param runnableName name of the runnable.
+   * @param logLevels map of the log level arguments to be saved.
+   */
+  public void setLogLevels(String runnableName, Map<String, LogEntry.Level> logLevels) {
+    Map<String, LogEntry.Level> newlogLevels = new ConcurrentHashMap<>(logLevels);
```
yaojiefeng: Yes. There will be no concurrent calls since this method is called in the wrap of `ContainerLock`. But immutable map cannot be used since we need to put values to the map if there is update

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -98,17 +109,23 @@ public Integer apply(BitSet input) {
   private final Condition containerChange;
   private final ZKClient zkClient;
   private final Multimap<String, ContainerInfo> containerStats;
+  private final Location applicationLocation;
+  private final List<String> allRunnableNames;
```
chtyim: Just call it `runnableNames`.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -98,17 +109,23 @@ public Integer apply(BitSet input) {
...
     completedContainerCount = Maps.newHashMap();
     startSequence = Lists.newLinkedList();
     containerLock = new ReentrantLock();
     containerChange = containerLock.newCondition();
-    resourceReport = new DefaultResourceReport(appId, appMasterResources);
+    resourceReport = new DefaultResourceReport(appId, appMasterResources, logLevels);
     zkClient = zookeeperClient;
     containerStats = HashMultimap.create();
+    applicationLocation = appLocation;
+    allRunnableNames = allRunnables;
```
chtyim: Better have the argument and the field named the same for clarity.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -132,16 +149,22 @@ void start(String runnableName, ContainerInfo containerInfo, TwillContainerLaunc
     try {
       int instanceId = getStartInstanceId(runnableName);
       RunId runId = getRunId(runnableName, instanceId);
+      Map<String, LogEntry.Level> logLevelsRunnable = resourceReport.getLogLevels().containsKey(runnableName) ?
+        resourceReport.getLogLevels().get(runnableName) : new HashMap<String, LogEntry.Level>();
```
chtyim: Just call `get` without calling `containsKey` and check for `null`.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -132,16 +149,22 @@ void start(String runnableName, ContainerInfo containerInfo, TwillContainerLaunc
     try {
       int instanceId = getStartInstanceId(runnableName);
       RunId runId = getRunId(runnableName, instanceId);
+      Map<String, LogEntry.Level> logLevelsRunnable = resourceReport.getLogLevels().containsKey(runnableName) ?
```
chtyim: The variable should be called `runnableLogLevels`.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -132,16 +149,22 @@ void start(String runnableName, ContainerInfo containerInfo, TwillContainerLaunc
...
+        resourceReport.getLogLevels().get(runnableName) : new HashMap<String, LogEntry.Level>();
+      logLevel = logLevelsRunnable.containsKey(Logger.ROOT_LOGGER_NAME) ?
```
chtyim: Is `ROOT_LOGGER_NAME` a special logger name? If that's the case, we should be consistently using the map only. I see some other classes having both `rootLogLevel` and the map.

twill-api/src/main/java/org/apache/twill/api/TwillPreparer.java
```diff
@@ -227,7 +227,7 @@
   TwillPreparer addSecureStore(SecureStore secureStore);
 
   /**
-   * Set the log level for Twill applications running in a container.
+   * Set the root log level for Twill applications in all containers.
```
chtyim: Same as above, do we have a way to set the root log level for a particular runnablE?

twill-api/src/main/java/org/apache/twill/api/TwillRunResources.java
```diff
@@ -59,6 +61,16 @@
   /**
    * @return the enabled log level for the container where the runnable is running in.
    */
+  @Deprecated
```
chtyim: Why deprecating a method, also add a `@deprecated` in the javadoc section to tell what's the alternative. E.g.

```
/**
 * ....
 * @deprecated Use {@link #getRootLogLevel()} instead.
 */

File: twill-api/src/main/java/org/apache/twill/internal/DefaultTwillRunResources.java [hunk 38]
```diff
@@ -129,7 +152,7 @@ public String toString() {
       ", memoryMB=" + memoryMB +
       ", host='" + host + '\'' +
       ", debugPort=" + debugPort +
-      ", logLevel=" + logLevel +
+      ", rootLogLevel=" + getRootLogLevel() +
```
chtyim: Use the `logLevels` field instead of calling the get method to be consistent.

twill-api/src/main/java/org/apache/twill/internal/DefaultTwillRunResources.java
```diff
@@ -90,8 +103,18 @@ public Integer getDebugPort() {
...
-    return logLevel;
+    return getRootLogLevel();
+  }
+
+  @Override
+  public Level getRootLogLevel() {
+    return getLogLevels().get(Logger.ROOT_LOGGER_NAME);
+  }
+
+  public Map<String, LogEntry.Level> getLogLevels() {
```
chtyim: Add `@Override`.

twill-core/src/main/java/org/apache/twill/internal/DefaultResourceReport.java
```diff
@@ -39,25 +44,33 @@
...
-    this.usedResources = HashMultimap.create();
+    this.usedResources = Multimaps.synchronizedSetMultimap(HashMultimap.<String, TwillRunResources>create());
     for (Map.Entry<String, Collection<TwillRunResources>> entry : resources.entrySet()) {
       this.usedResources.putAll(entry.getKey(), entry.getValue());
     }
     this.services = new AtomicReference<>(services);
+    this.logLevels = new ConcurrentHashMap<>();
+  }
+
+  public synchronized void setRunnableLogLevels(String runnableName, Map<String, LogEntry.Level> runnableLogLevels) {
```
chtyim: better name it as `updateRunnableLogLevels`.

chtyim: Also, maybe good to just return if the `runnableLogLevels` map is empty.

twill-core/src/main/java/org/apache/twill/internal/json/TwillRunResourcesCodec.java
```diff
@@ -55,9 +58,11 @@ public JsonElement serialize(TwillRunResources src, Type typeOfSrc, JsonSerializ
     if (src.getDebugPort() != null) {
       json.addProperty(DEBUG_PORT, src.getDebugPort());
     }
-    if (src.getLogLevel() != null) {
-      json.addProperty(LOG_LEVEL, src.getLogLevel().toString());
+    if (src.getRootLogLevel() != null) {
+      json.addProperty(ROOT_LOG_LEVEL, src.getRootLogLevel().name());
     }
+    json.add(LOG_LEVELS, context.serialize(src.getLogLevels(),
+                                           new TypeToken<Map<String, String>>() { }.getType()));
```
chtyim: The type should be `Map<String, LogEntry.Level>`.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,44 @@
...
+  Future<LogEntry.Level> setRootLogLevel(String runnableName, LogEntry.Level logLevel);
+
+  /**
+   * Set the log levels for requested logger names for Twill applications running in a container.
+   *
+   * @param logLevels The {@link Map} contains the requested logger names and log levels that need to be set.
+   * @return A {@link Future} that will be completed when the set log level operation has been done. It will carry the
+   *         {@link Map} of log levels as the result.
+   */
+  Future<Map<String, LogEntry.Level>> setLogLevels(Map<String, LogEntry.Level> logLevels);
```
yaojiefeng: Yes, you can set the root log level by calling like:
`setLogLevels(ImmutableMap.of("ROOT", TRACE))`

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,44 @@
...
+   * Set the log levels for requested logger names for a {@link TwillRunnable}.
+   *
+   * @param runnableName The name of the runnable to set the log level.
+   * @param logLevelsForRunnable The {@link Map} contains the requested logger name and log level that
+   *                             need to be changed.
+   * @return A {@link Future} that will be completed when the set log level operation has been done. It will carry the
+   *         {@link Map} of log levels as the result.
+   */
+  Future<Map<String, LogEntry.Level>> setLogLevels(String runnableName,
+                                                   Map<String, LogEntry.Level> logLevelsForRunnable);
```
yaojiefeng: For this PR, we do not have that functionality. You will have to change the log level of the particular logger name again to have the log level back.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -89,8 +91,7 @@ public String getRmSchedulerAddr() {
     return rmSchedulerAddr;
   }
 
-  @Nullable
-  public LogEntry.Level getLogLevel() {
-    return logLevel;
+  public Map<String, Map<String, LogEntry.Level>> getLogLevels() {
+    return logLevels;
   }
```
yaojiefeng: `TwillRuntimeSpecification` was added by me in previous PR: #13 to carry the log level information in `TwillPreparer`. So I do not think there will be compatibility issue.

twill-core/src/main/java/org/apache/twill/internal/json/ResourceReportAdapter.java
```diff
@@ -59,4 +72,61 @@ public ResourceReport fromJson(String json) {
...
+  /**
+   * A type adapter for serializing {@code Collection<TwillRunResources>} correctly
+   */
+  private static final class ResourceReportTypeAdapterFactory implements TypeAdapterFactory {
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+      if (!Collection.class.isAssignableFrom(type.getRawType())) {
+        return null;
```
yaojiefeng: This factory is specifically for serializing`Collection<TwillRunResources>` since Gson will not catch the inner class for it. So we only need to create the TypeAdapter when we encounter `Collection<TwillRunResources>`. Actually the implementation is simulated from `TwillRuntimeSpecificationAdapter`, which create factory to deal with similar situation, https://github.com/apache/twill/blob/master/twill-core/src/main/java/org/apache/twill/internal/json/TwillRuntimeSpecificationAdapter.java#L106

twill-core/src/main/java/org/apache/twill/internal/DefaultResourceReport.java
```diff
@@ -158,4 +174,31 @@ public String toString() {
       ", services=" + services +
       '}';
   }
+
+  private TwillRunResources wrapTwillRunResources(final String runnableName, TwillRunResources resources) {
```
yaojiefeng: The `logLevels` for all running containers are stored in `ResourceReport`, and `ResourceReport` have a map containing the `TwillRunResources` for each running runnable. We want to return the correct log levels for each runnable. So we wrap the `TwillRunResources` with `ForwardingTwillRunResources`, which has a delegation and we override the `getLogLevels` methods to get the log levels from the map in `ResourceReport`.

twill-core/src/main/java/org/apache/twill/internal/json/ResourceReportAdapter.java
```diff
@@ -59,4 +72,61 @@ public ResourceReport fromJson(String json) {
...
+        }
+
+        @Override
+        public Collection<V> read(JsonReader reader) throws IOException {
+          if (reader.peek() == JsonToken.NULL) {
+            reader.nextNull();
+            return null;
+          }
+          if (reader.peek() != JsonToken.BEGIN_ARRAY) {
+            return null;
```
yaojiefeng: Same as above, we only need to generate result when we encounter `Collection<TwillRunResources>`

twill-core/src/main/java/org/apache/twill/internal/json/ResourceReportAdapter.java
```diff
@@ -59,4 +72,61 @@ public ResourceReport fromJson(String json) {
...
+          }
+          writer.beginArray();
+          for (V value : collection) {
+            valueAdapter.write(writer, value);
+          }
+          writer.endArray();
+        }
+
+        @Override
+        public Collection<V> read(JsonReader reader) throws IOException {
```
yaojiefeng: Added

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -475,6 +496,11 @@ private void sendMessage(final String runnableName, final Message message,
     Futures.addCallback(controller.sendMessage(message), new FutureCallback<Message>() {
       @Override
       public void onSuccess(Message result) {
+        //update log level information
```
anew: onSuccess here really means that the message has been sent to the runnable successfully, right? But it does not reflect the actual state of the runnable itself? It that on purpose?

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -575,18 +601,74 @@ private boolean removeContainerInfo(String containerId) {
...
+
+    Map<String, LogEntry.Level> runnableLogLevels = convertLogLevelValuesToLogEntry(message.getCommand().getOptions());
+    if (!logLevels.containsKey(runnableName)) {
+      logLevels.put(runnableName, runnableLogLevels);
+    } else {
+      logLevels.get(runnableName).putAll(runnableLogLevels);
+    }
+  }
+
+  private Location saveLogLevels() {
```
anew: Can you explain why the log levels have to be saved to a file? It is not clear to me.

twill-yarn/src/main/java/org/apache/twill/internal/container/TwillContainerService.java
```diff
@@ -123,6 +142,9 @@ protected void doStart() throws Exception {
 
     runnable = Instances.newInstance((Class<TwillRunnable>) runnableClass);
     runnable.initialize(context);
+    if (!setLogLevels(Maps.transformValues(logLevels, Functions.toStringFunction()))) {
+      LOG.error("LoggerFactory is not a logback LoggerContext, cannot change log level");
```
anew: seems that the error logging belongs into the setLogLevels() method.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -296,8 +298,42 @@ public TwillPreparer addSecureStore(SecureStore secureStore) {
...
+      this.logLevel = logLevels.get(Logger.ROOT_LOGGER_NAME);
+    }
+    for (String runnableName : twillSpec.getRunnables().keySet()) {
+      saveLogLevels(runnableName, logLevels);
+    }
+    return this;
+  }
+
+  @Override
+  public TwillPreparer setLogLevels(String runnableName, Map<String, LogEntry.Level> runnableLogLeves) {
```
anew: typo: runnableLogLevels

twill-yarn/src/test/java/org/apache/twill/yarn/LogLevelChangeTestRun.java
```diff
@@ -0,0 +1,249 @@
...
+    controller.onRunning(new Runnable() {
+      @Override
+      public void run() {
+        running.countDown();
+      }
+    }, Threads.SAME_THREAD_EXECUTOR);
+    Assert.assertTrue(running.await(200, TimeUnit.SECONDS));
+
+    // assert that log level is DEBUG
+    waitForLogLevel(controller, LogLevelTestRunnable.class.getSimpleName(),
```
anew: it seems that all you are testing here is that the log level in the resources have changed. That, however, represents the log level that was sent to the runnable, but we don't know whether it was actually applied. To truly validate, the runnable should emit log messages, and the test should validate that after changing the log level, the level of messages that actually get collected has changed.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
anew: ok

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
yaojiefeng: We save the log levels to file because we have no way to know if there is message about log level change come before the launch of some containers(this can happen when we change instance count or restart the container). So when a message about log level change come, we will have it stored in the Map in `RunningContainers`. And if there is a container launch afterwards, we will check if there is update in the log levels, and localize it for the container if necessary.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
yaojiefeng: onSuccess here means that the message has been processed by the TwillContainerService, and the log levels have been set up. If there is some error setting up the log levels, an failed future with exception will be returned to callback, so it will go to onFailure() method

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
chtyim: We don't have to since it is not in twill-api

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,44 @@
    * @return A {@link Future} that will be completed when the restart operation has been done.
    */
   Future<String> restartInstances(String runnable, int instanceId, int... moreInstanceIds);
+
```
yaojiefeng: By docs on https://github.com/apache/twill/blob/master/twill-core/src/main/java/org/apache/twill/internal/AbstractZKServiceController.java#L107, that means the message was successfully processed.

twill-yarn/src/test/java/org/apache/twill/yarn/LogLevelChangeTestRun.java
yaojiefeng: Actually I am not quite sure how to collect logs emitted by runnable in unit test. So in runnable I have checked the log level change: https://github.com/apache/twill/pull/14/files#diff-e7b3cb437b83c0f34231004947a7ff7eR61, which is simulated from the `LogLevelTestRun.class`, https://github.com/apache/twill/blob/master/twill-yarn/src/test/java/org/apache/twill/yarn/LogLevelTestRun.java#L60

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,59 @@
...
+   */
+  Future<Map<String, LogEntry.Level>> setLogLevels(String runnableName,
+                                                   Map<String, LogEntry.Level> logLevelsForRunnable);
+
+  /**
+   * Reset the log levels of all runnables to the default log level, {@code LogEntry.Level.INFO}.
+   *
+   * @return A {@link Future} that will be completed when the set log level operation has been done.
+   */
+  Future<String> resetLogLevelsToDefault();
```
chtyim: Just call it `resetLogLevels`

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,59 @@
...
+
+  /**
+   * Reset the log levels of the given runnable to the default log level, {@code LogEntry.Level.INFO}.
+   *
+   * @return A {@link Future} that will be completed when the set log level operation has been done. It will carry the
+   *         runnable name as the result.
+   */
+  Future<String> resetRunnableLogLevelsToDefault(String runnableName);
```
chtyim: Same here, call it `resetLogLevels(String runnableName)` to mirror the `set` methods.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
chtyim: In that case, why we have the `setRootLogLevel` methods above? I don't like have two ways of doing the same task.

twill-api/src/main/java/org/apache/twill/api/TwillPreparer.java
```diff
@@ -227,15 +227,52 @@
...
+  TwillPreparer setRootLogLevel(LogEntry.Level logLevel);
+
+  /**
+   * Set the root log level for a particular runnable
+   *
+   * @param runnableName The name of the runnable to set the log level.
+   * @param logLevel the {@link LogEntry.Level} that should be set.
+   * @return This {@link TwillPreparer}.
+   */
+  TwillPreparer setRootLogLevel(String runnableName, LogEntry.Level logLevel);
```
chtyim: I prefer not to introduce the `setRootLogLevel` and the deprecated `setLogLevel` replaced with call to `setLogLevel(Map)` with logger name `Logger.ROOT_LOGGER_NAME` to set root logger log level.

twill-api/src/main/java/org/apache/twill/api/TwillRunResources.java
```diff
@@ -58,7 +60,18 @@
...
    * @return the enabled log level for the container where the runnable is running in.
+   * @deprecated Use {@link #getRootLogLevel()} instead.
    */
+  @Deprecated
   LogEntry.Level getLogLevel();
 
+  /**
+   * @return the enabled log level for the container where the runnable is running in.
+   */
+  LogEntry.Level getRootLogLevel();
```
chtyim: Same reason as above, don't introduce this method just specialized for root logger.

twill-api/src/main/java/org/apache/twill/internal/DefaultTwillRunResources.java
```diff
@@ -30,17 +35,25 @@
...
   private final Integer debugPort;
-  private final Level logLevel;
+  private final Map<String, LogEntry.Level> logLevels;
+
+  /**
+   * Constructor to create an instance of {@link DefaultTwillRunResources} with empty log levels.
+   */
+  public DefaultTwillRunResources(int instanceId, String containerId, int cores, int memoryMB,
+                                  String host, Integer debugPort) {
+    this(instanceId, containerId, cores, memoryMB, host, debugPort, new HashMap<String, Level>());
```
chtyim: Use `Collections.<String, Level>emptyMap()` instead.

twill-api/src/main/java/org/apache/twill/internal/DefaultTwillRunResources.java
```diff
@@ -129,7 +153,7 @@ public String toString() {
       ", memoryMB=" + memoryMB +
       ", host='" + host + '\'' +
       ", debugPort=" + debugPort +
-      ", logLevel=" + logLevel +
+      ", rootLogLevel=" + logLevels.get(Logger.ROOT_LOGGER_NAME) +
```
chtyim: Why only has root log level? The `toString` should includes the whole `logLevels` map.

twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
```diff
@@ -170,6 +171,37 @@ public String apply(Set<String> input) {
...
+
+  @Override
+  public Future<Map<String, LogEntry.Level>> setLogLevels(String runnableName,
+                                                          Map<String, LogEntry.Level> runnableLogLevels) {
+    return sendMessage(SystemMessages.setLogLevels(runnableName, runnableLogLevels), runnableLogLevels);
+  }
+
+  @Override
+  public Future<String> resetLogLevelsToDefault() {
+    return sendMessage(SystemMessages.resetLogLevels(null), "all");
```
chtyim: This is inconsistent with the `setLogLevels` method, which doesn't require explicitly passing in `null` for all runnables case.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
chtyim: Also, this reset everything? So there is no way to reset only a subset of loggers?

twill-core/src/main/java/org/apache/twill/internal/ContainerLiveNodeData.java
```diff
@@ -43,4 +51,15 @@ public String getHost() {
   public String getDebugPort() {
     return debugPort;
   }
+
+  public Map<String, LogEntry.Level> getLogLevels() {
+    return logLevels;
+  }
+
+  public void updateLogLevels(Map<String, LogEntry.Level> logLevels, boolean isReset) {
+    if (isReset) {
```
chtyim: This is not a good way overloading the responsibility of this method. Better have a separate method for resetting (and that mirrors the actual functionality being exposed through the public API).

twill-core/src/main/java/org/apache/twill/internal/DefaultResourceReport.java
```diff
@@ -41,19 +46,15 @@
...
+    this(applicationId, masterResources, Collections.<String, Collection<TwillRunResources>>emptyMap(),
+         Collections.<String>emptyList());
   }
 
   public DefaultResourceReport(String applicationId, TwillRunResources masterResources,
                                Map<String, Collection<TwillRunResources>> resources, List<String> services) {
     this.applicationId = applicationId;
     this.appMasterResources = masterResources;
-    this.usedResources = HashMultimap.create();
+    this.usedResources = Multimaps.synchronizedSetMultimap(HashMultimap.<String, TwillRunResources>create());
```
chtyim: Is this change still necessary?

twill-core/src/main/java/org/apache/twill/internal/DefaultResourceReport.java
```diff
@@ -17,17 +17,22 @@
...
 import org.apache.twill.api.ResourceReport;
 import org.apache.twill.api.TwillRunResources;
+import org.apache.twill.api.logging.LogEntry;
 
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
```
chtyim: unused import?

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -17,12 +17,14 @@
  */
 package org.apache.twill.internal;
 
+import ch.qos.logback.classic.Logger;
```
chtyim: Unused import?

twill-yarn/src/main/java/org/apache/twill/internal/ServiceMain.java
```diff
@@ -264,6 +262,10 @@ protected String getLoggerLevel(Logger logger) {
     return "OFF";
   }
 
+  protected String getLogLevels() {
+    return "";
```
chtyim: Why we can't set the custom log levels at launch in the same programmatic way as when a system message is received?

twill-api/src/main/java/org/apache/twill/api/TwillController.java
yaojiefeng: `resetLogLevels` will reset all log levels passed in runtime except the root to null. If user want to reset log level for a particular logger, he can use `setLogLevels(ImmutableMap.of(someloggerName, null))` to set the log level since null is actually a valid log level for log back

twill-yarn/src/main/java/org/apache/twill/internal/ServiceMain.java
yaojiefeng: For a launch, the `TwillContainerService` is constructed before the run of `doMain` method, which means the we have not called `configureLogger()` method at that time so log level setting at constructor will not work.

twill-core/src/main/java/org/apache/twill/internal/json/TwillRunResourcesCodec.java
```diff
@@ -66,13 +67,14 @@ public JsonElement serialize(TwillRunResources src, Type typeOfSrc, JsonSerializ
   public TwillRunResources deserialize(JsonElement json, Type typeOfT,
                                            JsonDeserializationContext context) throws JsonParseException {
     JsonObject jsonObj = json.getAsJsonObject();
+    Map<String, LogEntry.Level> logLevels =
+      context.deserialize(jsonObj.get("logLevels"), new TypeToken<Map<String, LogEntry.Level>>() { }.getType());
```
chtyim: Since constants are defined, we should use those constants in deserialization as well.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -575,18 +597,63 @@ private boolean removeContainerInfo(String containerId) {
...
+    LOG.debug("save the log level file");
+    try {
+      Gson gson = new GsonBuilder().serializeNulls().create();
+      String jsonStr = gson.toJson(logLevels);
+      String fileName = Hashing.md5().hashString(jsonStr) + "." + Constants.Files.LOG_LEVELS;
+      Location location = applicationLocation.append(fileName);
+      if (!location.exists()) {
+        try (Writer writer = new OutputStreamWriter(location.getOutputStream(), Charsets.UTF_8)) {
+          gson.toJson(logLevels, new TypeToken<Map<String, Map<String, LogEntry.Level>>>() { }.getType(), writer);
```
chtyim: Why need to encode again? You can just write the `jsonStr` to the file.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -605,5 +672,14 @@ public synchronized Integer getDebugPort() {
...
       return dynamicDebugPort;
     }
+
+    @Override
+    public synchronized Map<String, LogEntry.Level> getLogLevels() {
+      ContainerLiveNodeData liveData = controller.getLiveNodeData();
+      if (liveData != null) {
+        return liveData.getLogLevels();
+      }
+      return new HashMap<>();
```
chtyim: return `Collections.emptyMap()`.

twill-yarn/src/main/java/org/apache/twill/internal/container/TwillContainerService.java
```diff
@@ -56,24 +66,27 @@
...
+                                                   Map<String, LogEntry.Level> logLevels) {
     // if debugging is enabled, log the port and register it in service discovery.
     String debugPort = System.getProperty("twill.debug.port");
     if (debugPort != null) {
       LOG.info("JVM is listening for debugger on port {}", debugPort);
     }
     return new ContainerLiveNodeData(containerInfo.getId(),
                                      containerInfo.getHost().getCanonicalHostName(),
-                                     debugPort);
+                                     debugPort, logLevels);
```
chtyim: I think we should filter out the one that has `null` value, since those are representing reset.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,30 @@
    * @return A {@link Future} that will be completed when the restart operation has been done.
    */
   Future<String> restartInstances(String runnable, int instanceId, int... moreInstanceIds);
+
+  /**
+   * Set the log levels for requested logger names for Twill applications running in a container.
+   * The log level for a logger name can be {@code null} except for the root logger, which will reset the log level for
```
yaojiefeng: We check it using `Preconditions.checkArgument`, so an exception will be thrown for null log level for root logger.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
hsaputra: where did you see the `setRootLogLevel` method definition?

twill-api/src/main/java/org/apache/twill/api/TwillController.java
anew: Still, the way I understand this is that it means, the message was received by the app master and passed on to the runnables. We have no idea whether the runnables have received or processed it. That can only be determined by examining the live node info for each runnable.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,45 @@
...
+
+  /**
+   * Reset the log levels of all runnables to the default log level, {@code LogEntry.Level.INFO}.
+   *
+   * @return A {@link Future} that will be completed when the set log level operation has been done.
+   */
+  Future<String> resetLogLevels();
+
+  /**
+   * Reset the log levels of the given runnable to the log levels when .
```
anew: this sentence is incomplete. Does this mean restore the log level to what it was before the last setLogLevel()? Or before setLogLevel was called for the first time? Or to what it was in the original spec?

twill-yarn/src/main/java/org/apache/twill/internal/container/TwillContainerService.java
```diff
@@ -51,29 +63,36 @@
...
-                               Location applicationLocation) {
+                               Location applicationLocation, Map<String, LogEntry.Level> defaultLogLevels,
+                               Map<String, LogEntry.Level> logLevels) {
     super(zkClient, runId, applicationLocation);
 
     this.specification = specification;
     this.classLoader = classLoader;
-    this.containerLiveNodeData = createLiveNodeData(containerInfo);
+    this.defaultLogLevels = ImmutableMap.copyOf(defaultLogLevels);
+    this.oldLogLevels = Collections.synchronizedMap(new HashMap<>(defaultLogLevels));
```
anew: can you explain what oldLogLevels is for? My understanding is this:
- when a log level is updated, you save previous level in oldLogLevels
- when the log level is reset, you restore the log level from oldLogLevels
Is that correct?

If so, then consider this scenario:
- root log level is INFO
- I debug a problem, so I set the root log level to DEBUG
- I set the log level for one logger to TRACE
- now I found the problem. 
- I reset the log level for that particular logger
- then I set the root log level back to INFO

what will be the log level for that particular logger after all this? INFO or DEBUG?

twill-api/src/main/java/org/apache/twill/api/TwillController.java
yaojiefeng: My bad for not updating the comments. The `resetLogLevels` will reset the log level of a runnable to the time when it starts up. This means it will reset the log level to what `logback.xml` is + the log level requested from `TwillPreparer`

twill-yarn/src/main/java/org/apache/twill/internal/container/TwillContainerService.java
yaojiefeng: The `defaultLogLevels` are the log levels passed with `TwillPreparer`, when resetting the log levels, we will need it to update the live node info to indicate the log levels. The `oldLogLevels` is the same as the `defaultLogLevels` when the runnable starts up. We will not update it for each log level change. It is used to save old log levels for loggers for the first time change as we do not know what is in the `logback.xml`. When reset, we reset based on this map.

yaojiefeng: When you reset, you cannot reset the log level for a particular logger. It will reset the log levels for all loggers which are changed during runtime. So if you call reset, it will reset the log level of ROOT to INFO, and the particular logger to the log level when the runnable starts. If it was null initially, its log level will be following the ROOT log level, INFO

anew: I see. Seems ok.

anew:
OK, my concerns are addressed and it LGTM. Please wait for @chtyim to do another pass before this gets merged.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
chtyim: @hsaputra I think it get removed after this comment, but this comment sticks.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,48 @@
...
+   * @return A {@link Future} that will be completed when the set log level operation has been done. It will carry the
+   *         {@link Map} of log levels as the result.
+   */
+  Future<Map<String, LogEntry.Level>> setLogLevels(String runnableName,
+                                                   Map<String, LogEntry.Level> logLevelsForRunnable);
+
+  /**
+   * Reset the log levels of all runnables.
+   * The log levels will be the same as when the runnables start up.
+   *
```
chtyim: Missed the `@param loggerNames` doc.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,48 @@
...
+   * @return A {@link Future} that will be completed when the set log level operation has been done. The future result
+   *         is the logger names provided in the parameter.
+   */
+  Future<String[]> resetLogLevels(String...loggerNames);
+
+  /**
+   * Reset the log levels of the given runnable.
+   * The log levels will be same as when the runnable starts up.
+   *
```
chtyim: Missed the `@param loggerNames` doc.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -384,6 +408,24 @@ private void setEnv(String runnableName, Map<String, String> env, boolean overwr
     }
   }
 
+  private void saveLogLevels(LogEntry.Level level) {
+    level = level == null ? LogEntry.Level.INFO : level;
```
chtyim: Shouldn't default it to `INFO` at all. It should be treated the same as `null` is provided as the log level value as in the other `saveLogLevels` method, which will throw `IllegalArgumentException`. Also, rename the `level` parameter to `rootLogLevel` to be clear.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -151,8 +152,9 @@
     this.reservedMemory = yarnConfig.getInt(Configs.Keys.JAVA_RESERVED_MEMORY_MB,
                                             Configs.Defaults.JAVA_RESERVED_MEMORY_MB);
     this.extraOptions = extraOptions;
-    this.logLevel = logLevel;
     this.classAcceptor = new ClassAcceptor();
+    this.logLevel = logLevel;
```
chtyim: Shouldn't have this field anymore. Every log level info should be in the `logLevels` map

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -384,6 +408,24 @@ private void setEnv(String runnableName, Map<String, String> env, boolean overwr
...
+    }
+  }
+
+  private void saveLogLevels(String runnableName, Map<String, LogEntry.Level> logLevels) {
+    Map<String, String> newLevels = new HashMap<>();
+    for (Map.Entry<String, LogEntry.Level> entry : logLevels.entrySet()) {
+      Preconditions.checkArgument(entry.getValue() != null, "Log level cannot be null for logger {}", entry.getKey());
+      newLevels.put(entry.getKey(), entry.getValue().name());
+    }
+    this.logLevels.get(runnableName).putAll(newLevels);
```
chtyim: We should always create a new map instead of calling `putAll`. The reason is the exposed method is call `setLogLevels`, not `addLogLevels` or `appendLogLevels`, which implies a overwriting instead of appending.

twill-api/src/main/java/org/apache/twill/api/TwillPreparer.java
```diff
@@ -227,15 +227,36 @@
...
    *                 The level match the {@code Logback} levels.
    * @return This {@link TwillPreparer}.
+   * @deprecated Use {@link #setLogLevels(Map)} with key {@link org.slf4j.Logger#ROOT_LOGGER_NAME} instead.
    */
+  @Deprecated
   TwillPreparer setLogLevel(LogEntry.Level logLevel);
 
   /**
+   * Set the log levels for requested logger names for Twill applications running in a container. The log level whose
+   * value is null will be ignored
```
chtyim: I think the implementation actually throws `IllegalArgumentException`, which is better than silently ignored. Please update the doc here accordingly.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -89,4 +90,45 @@
...
+   */
+  Future<Map<String, LogEntry.Level>> setLogLevels(String runnableName,
+                                                   Map<String, LogEntry.Level> logLevelsForRunnable);
+
+  /**
+   * Reset the log levels of all runnables to the default log level, {@code LogEntry.Level.INFO}.
+   *
+   * @return A {@link Future} that will be completed when the set log level operation has been done.
+   */
+  Future<String> resetLogLevels();
```
yaojiefeng: Changed the value in `Future` to be the loggerNames passed through this methods. Empty means resetting for all loggers

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
chtyim: What I meant was, since the method is "set", it should be setting / overwriting, not appending. And I think it's ok for the user to gather and pass in the map.

chtyim:
LGTM

hsaputra:
Congrats @yaojiefeng ! =)

yaojiefeng:
Thanks! @hsaputra

PR #16, opened 2016-11-22
[TWILL-199] Return the offset to read next message in KafkaConsumer.MessageCallback#onReceive

maochf:
https://issues.apache.org/jira/browse/TWILL-199

### DIFF
--- a/twill-api/src/main/java/org/apache/twill/api/ServiceAnnouncer.java
+++ b/twill-api/src/main/java/org/apache/twill/api/ServiceAnnouncer.java
@@ -32,7 +32,7 @@ public interface ServiceAnnouncer {
   Cancellable announce(String serviceName, int port);
 
   /**
-   * Registers an endpoint that could be discovered by external party with a payload
+   * Registers an endpoint that could be discovered by external party with a payload.
    * @param serviceName Name of the endpoint
    * @param port Port of the endpoint
    * @param payload byte array payload
--- a/twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
@@ -229,9 +229,11 @@ public abstract class AbstractTwillController extends AbstractZKServiceControlle
     }
 
     @Override
-    public void onReceived(Iterator<FetchedMessage> messages) {
+    public long onReceived(Iterator<FetchedMessage> messages) {
+      long nextOffset = -1L;
       while (messages.hasNext()) {
-        String json = Charsets.UTF_8.decode(messages.next().getPayload()).toString();
+        FetchedMessage message = messages.next();
+        String json = Charsets.UTF_8.decode(message.getPayload()).toString();
         try {
           LogEntry entry = GSON.fromJson(json, LogEntry.class);
           if (entry != null) {
@@ -240,7 +242,9 @@ public abstract class AbstractTwillController extends AbstractZKServiceControlle
         } catch (Exception e) {
           LOG.error("Failed to decode log entry {}", json, e);
         }
+        nextOffset = message.getNextOffset();
       }
+      return nextOffset;
     }
 
     @Override
--- a/twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
@@ -57,6 +57,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
@@ -273,17 +274,17 @@ final class SimpleKafkaConsumer implements KafkaConsumer {
       final AtomicBoolean stopped = new AtomicBoolean();
       return new MessageCallback() {
         @Override
-        public void onReceived(final Iterator<FetchedMessage> messages) {
+        public long onReceived(final Iterator<FetchedMessage> messages) {
           if (stopped.get()) {
-            return;
+            return -1L;
           }
-          Futures.getUnchecked(executor.submit(new Runnable() {
+          return Futures.getUnchecked(executor.submit(new Callable<Long>() {
             @Override
-            public void run() {
+            public Long call() {
               if (stopped.get()) {
-                return;
+                return -1L;
               }
-              callback.onReceived(messages);
+              return callback.onReceived(messages);
             }
           }));
         }
@@ -450,7 +451,7 @@ final class SimpleKafkaConsumer implements KafkaConsumer {
     private void invokeCallback(ByteBufferMessageSet messages, AtomicLong offset) {
       long savedOffset = offset.get();
       try {
-        callback.onReceived(createFetchedMessages(messages, offset));
+        offset.set(callback.onReceived(createFetchedMessages(messages, offset)));
       } catch (Throwable t) {
         LOG.error("Callback throws exception. Retry from offset {} for {}", startOffset, topicPart, t);
         offset.set(savedOffset);
@@ -475,9 +476,9 @@ final class SimpleKafkaConsumer implements KafkaConsumer {
               continue;
             }
 
-            offset.set(message.nextOffset());
             fetchedMessage.setPayload(message.message().payload());
-            fetchedMessage.setNextOffset(offset.get());
+            fetchedMessage.setOffset(message.offset());
+            fetchedMessage.setNextOffset(message.nextOffset());
 
             return fetchedMessage;
           }
--- a/twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
+++ b/twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
@@ -35,8 +35,9 @@ public interface KafkaConsumer {
      * Invoked when new messages is available.
      * @param messages Iterator of new messages. The {@link FetchedMessage} instance maybe reused in the Iterator
      *                 and across different invocation.
+     * @return The offset of the message to be fetched next.
      */
-    void onReceived(Iterator<FetchedMessage> messages);
+    long onReceived(Iterator<FetchedMessage> messages);
 
     /**
      * Invoked when message consumption is stopped. When this method is invoked,
--- a/twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
+++ b/twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
@@ -42,10 +42,8 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Iterator;
 import java.util.Properties;
-import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
@@ -89,8 +87,8 @@ public class KafkaTest {
   @Test
   public void testKafkaClientReconnect() throws Exception {
     String topic = "backoff";
-    Properties kafkServerConfig = generateKafkaConfig(zkServer.getConnectionStr() + "/backoff");
-    EmbeddedKafkaServer server = new EmbeddedKafkaServer(kafkServerConfig);
+    Properties kafkaServerConfig = generateKafkaConfig(zkServer.getConnectionStr() + "/backoff");
+    EmbeddedKafkaServer server = new EmbeddedKafkaServer(kafkaServerConfig);
 
     ZKClientService zkClient = ZKClientService.Builder.of(zkServer.getConnectionStr() + "/backoff").build();
     zkClient.startAndWait();
@@ -106,15 +104,19 @@ public class KafkaTest {
           // Publish a messages
           createPublishThread(kafkaClient, topic, Compression.NONE, "First message", 1).start();
 
-          // Creater a consumer
+          // Create a consumer
           final BlockingQueue<String> queue = new LinkedBlockingQueue<>();
           Cancellable cancel = kafkaClient.getConsumer().prepare().add(topic, 0, 0)
             .consume(new KafkaConsumer.MessageCallback() {
               @Override
-              public void onReceived(Iterator<FetchedMessage> messages) {
+              public long onReceived(Iterator<FetchedMessage> messages) {
+                long nextOffset = -1L;
                 while (messages.hasNext()) {
-                  queue.offer(Charsets.UTF_8.decode(messages.next().getPayload()).toString());
+                  FetchedMessage message = messages.next();
+                  nextOffset = message.getNextOffset();
+                  queue.offer(Charsets.UTF_8.decode(message.getPayload()).toString());
                 }
+                return nextOffset;
               }
 
               @Override
@@ -130,7 +132,7 @@ public class KafkaTest {
 
           // Start the server again.
           // Needs to create a new instance with the same config since guava service cannot be restarted
-          server = new EmbeddedKafkaServer(kafkServerConfig);
+          server = new EmbeddedKafkaServer(kafkaServerConfig);
           server.startAndWait();
 
           // Publish another message
@@ -170,11 +172,15 @@ public class KafkaTest {
     Cancellable cancel = kafkaClient.getConsumer().prepare().add(topic, 0, 0).consume(new KafkaConsumer
       .MessageCallback() {
       @Override
-      public void onReceived(Iterator<FetchedMessage> messages) {
+      public long onReceived(Iterator<FetchedMessage> messages) {
+        long nextOffset = -1;
         while (messages.hasNext()) {
-          LOG.info(Charsets.UTF_8.decode(messages.next().getPayload()).toString());
+          FetchedMessage message = messages.next();
+          nextOffset = message.getNextOffset();
+          LOG.info(Charsets.UTF_8.decode(message.getPayload()).toString());
           latch.countDown();
         }
+        return nextOffset;
       }
 
       @Override
@@ -184,6 +190,51 @@ public class KafkaTest {
     });
 
     Assert.assertTrue(latch.await(5, TimeUnit.SECONDS));
+    cancel.cancel();
+    Assert.assertTrue(stopLatch.await(1, TimeUnit.SECONDS));
+  }
+
+  @Test
+  public void testKafkaClientSkipNext() throws Exception {
+    String topic = "testClientSkipNext";
+    // Publish 30 messages with indecies the same as offsets within the range 0 - 29
+    Thread t1 = createPublishThread(kafkaClient, topic, Compression.GZIP, "GZIP Testing message", 10);
+    t1.start();
+    t1.join();
+    Thread t2 = createPublishThread(kafkaClient, topic, Compression.NONE, "Testing message", 10, 10);
+    t2.start();
+    t2.join();
+    Thread t3 = createPublishThread(kafkaClient, topic, Compression.SNAPPY, "Snappy Testing message", 10, 20);
+    t3.start();
+    t3.join();
+
+    final CountDownLatch stopLatch = new CountDownLatch(1);
+    final BlockingQueue<Long> offsetQueue = new LinkedBlockingQueue<>();
+    Cancellable cancel = kafkaClient.getConsumer().prepare().add(topic, 0, 0).consume(
+      new KafkaConsumer.MessageCallback() {
+      @Override
+      public long onReceived(Iterator<FetchedMessage> messages) {
+        long nextOffset = -1L;
+        if (messages.hasNext()) {
+          FetchedMessage message = messages.next();
+          nextOffset = message.getNextOffset() + 1;
+          offsetQueue.offer(message.getOffset());
+          LOG.info(Charsets.UTF_8.decode(message.getPayload()).toString());
+          return nextOffset;
+        }
+        return nextOffset;
+      }
+
+      @Override
+      public void finished() {
+        stopLatch.countDown();
+      }
+    });
+    // 15 messages should be in the queue since onReceived returns `message.getNextOffset() + 1` as next offset to read
+    for (long i = 0; i < 30; i += 2) {
+      Assert.assertEquals(i, (long) offsetQueue.poll(60, TimeUnit.SECONDS));
+    }
+    Assert.assertNull(offsetQueue.poll(2, TimeUnit.SECONDS));
     cancel.cancel();
     Assert.assertTrue(stopLatch.await(1, TimeUnit.SECONDS));
   }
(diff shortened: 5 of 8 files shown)


twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
```diff
@@ -35,8 +35,10 @@
      * Invoked when new messages is available.
      * @param messages Iterator of new messages. The {@link FetchedMessage} instance maybe reused in the Iterator
      *                 and across different invocation.
+     * @return A long larger than zero as the offset to restart fetching messages when offset error is caught,
+     *         {@code -1} if the error cannot be resolved. Returns {@code 0} if no need to restart fetching.
      */
-    void onReceived(Iterator<FetchedMessage> messages);
+    long onReceived(Iterator<FetchedMessage> messages);
```
chtyim: How to indicate to restart from earliest or latest?

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaOffsetProvider.java
```diff
@@ -0,0 +1,36 @@
...
+package org.apache.twill.kafka.client;
+
+/**
+ * Define interface that could provide a method to check whether a message meets a given condition. If the condition is
+ * not met, the method will return the next offset to continue searching for the message meeting this condition.
+ */
+public interface KafkaOffsetProvider {
+
+  /**
+   * Check whether a message meets a given condition. If the condition is not met, return the next offset to
```
chtyim: I don't see any condition. How is the condition provided?

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaOffsetProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.twill.kafka.client;
+
+/**
+ * Define interface that could provide a method to check whether a message meets a given condition. If the condition is
+ * not met, the method will return the next offset to continue searching for the message meeting this condition.
+ */
+public interface KafkaOffsetProvider {
```
chtyim: How is this interface being used? I don't see any usage of this class.

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
maochf: How about -1 for latest and -2 for earliest? and < -2 for error?

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaOffsetProvider.java
maochf: It will be provided in the constructor. For instance, the timestamp will be passed as a parameter in the constructor. Other parameters can also be passed instead, and they can be used in `getCandidateOffset` method

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaOffsetProvider.java
maochf: It will be implemented outside of Twill. The objects will be passed to the constructor of `MessageCallback`

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaOffsetProvider.java
chtyim: It seems like this is not a clean contract. This is an interface, why we have to assume something passed in the constructor of the implementation class? Seem very confusing.

maochf: Right, how about changing this to an abstract class instead?

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
albertshau: -1 for latest and -2 for earliest seems to make sense. 0 is a valid kafka offset, though I suppose practically speaking, any use of 0 could be replaced by -2.

It is unclear what an 'error' means. Does it mean the consumer should stop consuming? I think we can just leave that out.

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
```diff
@@ -35,8 +35,10 @@
      * Invoked when new messages is available.
      * @param messages Iterator of new messages. The {@link FetchedMessage} instance maybe reused in the Iterator
      *                 and across different invocation.
+     * @return A long larger than zero as the offset to restart fetching messages when offset error is caught,
```
albertshau: what happens if the offset returned is out of bounds? Should document the expected behavior in such circumstances. Looking at the implementation, it seems like we read from the earliest offset.

twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
```diff
@@ -206,9 +206,12 @@ private LogMessageCallback(Iterable<LogHandler> logHandlers) {
     }
 
     @Override
-    public void onReceived(Iterator<FetchedMessage> messages) {
+    public long onReceived(Iterator<FetchedMessage> messages) {
+      long nextOffset = Long.MIN_VALUE;
```
chtyim: Why it is min value?

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
```diff
@@ -68,7 +69,7 @@
 /**
  * A {@link KafkaConsumer} implementation using the scala kafka api.
  */
-final class SimpleKafkaConsumer implements KafkaConsumer {
+public final class SimpleKafkaConsumer implements KafkaConsumer {
```
chtyim: Why turn this to public? This is an internal implementation of `KafkaConsumer`, which shouldn't be public

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
```diff
@@ -273,17 +274,19 @@ private MessageCallback wrapCallback(final MessageCallback callback,
       final AtomicBoolean stopped = new AtomicBoolean();
       return new MessageCallback() {
         @Override
-        public void onReceived(final Iterator<FetchedMessage> messages) {
+        public long onReceived(final Iterator<FetchedMessage> messages) {
           if (stopped.get()) {
-            return;
+            return Long.MIN_VALUE;
```
chtyim: Why min value?

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
```diff
@@ -273,17 +274,19 @@ private MessageCallback wrapCallback(final MessageCallback callback,
...
+            long nextOffset = Long.MIN_VALUE;
             @Override
-            public void run() {
+            public Long call() {
               if (stopped.get()) {
-                return;
+                return nextOffset;
               }
-              callback.onReceived(messages);
+              nextOffset = callback.onReceived(messages);
```
chtyim: Just `return callback.onReceived(messages)`. No need to store it to a local variable.

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
```diff
@@ -450,7 +453,10 @@ private boolean sleepIfEmpty(ByteBufferMessageSet messages) {
     private void invokeCallback(ByteBufferMessageSet messages, AtomicLong offset) {
       long savedOffset = offset.get();
       try {
-        callback.onReceived(createFetchedMessages(messages, offset));
+        Long nextOffset = callback.onReceived(createFetchedMessages(messages, offset));
```
chtyim: Why use `Long`? Shouldn't it be `long`?

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
```diff
@@ -35,8 +35,10 @@
      * Invoked when new messages is available.
      * @param messages Iterator of new messages. The {@link FetchedMessage} instance maybe reused in the Iterator
      *                 and across different invocation.
+     * @return The offset of the message to be fetched next. Returns {@code Long.MIN_VALUE} to keep the current offset
```
chtyim: In what case is unchange?

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: Because in line 286, the stored `nextOffset` should be returned

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: I wanted to use the `getLastOffset` method in the test. Haven't figured out a way to extract the logic in `getLastOffset` to other places without exposing the `SimpleKafkaConsumer` class because `getLastOffset` uses a lot of private members in `SimpleKafkaConsumer`

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: In this case, no message is processed, so the offset should remain unchanged to start reading from the current offset again next time.

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
chtyim: That's not a good reason to turn this class to public. Why the unit test have to use this method? Can't the unit test just return some offset that it know of?

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
chtyim: Using special value is an anti-pattern. Since you are changing the `MessageCallback` API already, probably better to have the offset being used for the fetching call being passed to the `onReceived` method.

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
chtyim: From the current code logic, line 286 always return `Long.MIN_VALUE`.

twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
```diff
@@ -170,11 +174,128 @@ public void testKafkaClient() throws Exception {
...
+      public void finished() {
+        //no-op
+      }
+    });
+
+    long startOffset = offsetQueue.poll(360, TimeUnit.SECONDS);
+    initCancel.cancel();
+
+    Cancellable cancel = kafkaClient.getConsumer().prepare().add(topic, 0, startOffset).consume(new KafkaConsumer
+      .MessageCallback() {
```
chtyim: This is an awkward line break.

twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
```diff
@@ -170,11 +174,128 @@ public void testKafkaClient() throws Exception {
...
+    final CountDownLatch stopLatch = new CountDownLatch(1);
+    final BlockingQueue<Long> offsetQueue = new LinkedBlockingQueue<>();
+    // Creater a consumer
+    final SimpleKafkaConsumer consumer = (SimpleKafkaConsumer) kafkaClient.getConsumer();
+    Cancellable initCancel = kafkaClient.getConsumer().prepare().add(topic, 0, 0).consume(new KafkaConsumer
+      .MessageCallback() {
+      long minOffset = -2; // earliest msg
+      long maxOffset = -1; // latest msg
+      @Override
+      // Use binary search to find the offset of the message with the index matching startIdx. Returns the next offset
```
chtyim: Seems like this unit-test is unnecessarily complicated. All you want to test is that the offset being returned from the `onReceived` method is being honored, right?

chtyim:
There are checkstyle failure, please fix them as well.

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: Right, this can be done in the unit test. To phrase my question better, what can be a better way to perform the same logic as `getLastOffset` outside of `SimpleKafkaConsumer`? Maybe expose the parameter passed to `SimpleKafkaConsumer` constructor in `ZKKafkaClientService`?

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: If the offset is passed to `onReceived`, it seems that there's no need to return an offset? Because now the returned offset is used to set the offset outside of the `onReceived` method.

chtyim: Why? The offset passed to this method is the offset being used for fetching, hence the creation of the iterator. The offset returned, on the other hand governs the offset to use for the next fetch

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
chtyim: It's independent of this PR, isn't it?

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: Maybe to pass an `AtomicLong` to this method, and it contains the initial offset and can be set to the next offset? Just realized this is probably an implementation detail in `SimpleKafkaConsumer` and is not general enough to be done in the interface.

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: Yes, it's an implementation detail. Just wondering for the future work.

twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
maochf: Right, since the offsets and indices of the messages are known, it can directly jump to the desired idx without binary search.

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
chtyim: Using AtomicLong as a mutable parameter is not good. In general, immutability gives cleaner API

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
maochf: I was assuming that this `onReceived` can be called multiple times. Therefore, at line 286 when stopped is set to true, messages processed in previous calls of `onReceived` should be skipped.

twill-api/src/main/java/org/apache/twill/api/ServiceAnnouncer.java
```diff
@@ -32,7 +32,7 @@
   Cancellable announce(String serviceName, int port);
 
   /**
-   * Registers an endpoint that could be discovered by external party with a payload
+   * Registers an endpoint that could be discovered by external party with a payload.
```
maochf: checkstyle fix

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
```diff
@@ -35,12 +35,14 @@
      * Invoked when new messages is available.
      * @param messages Iterator of new messages. The {@link FetchedMessage} instance maybe reused in the Iterator
      *                 and across different invocation.
+     * @param startOffset Offset of the current message to be consumed.
+     * @return The offset of the message to be fetched next.
      */
-    void onReceived(Iterator<FetchedMessage> messages);
+    long onReceived(Iterator<FetchedMessage> messages, long startOffset);
```
chtyim: Can you revert the ordering of the parameters? Logically it fetches from the `startOffset` to produce the `Iterator`.

twill-core/src/main/java/org/apache/twill/internal/kafka/client/SimpleKafkaConsumer.java
```diff
@@ -273,17 +274,19 @@ private MessageCallback wrapCallback(final MessageCallback callback,
...
         @Override
-        public void onReceived(final Iterator<FetchedMessage> messages) {
+        public long onReceived(final Iterator<FetchedMessage> messages, final long startOffset) {
           if (stopped.get()) {
-            return;
+            return startOffset;
           }
-          Futures.getUnchecked(executor.submit(new Runnable() {
+          return Futures.getUnchecked(executor.submit(new Callable<Long>() {
+            long nextOffset = startOffset;
```
chtyim: You don't need this local variable. If the consumer is stopped, it always returns the `startOffset` inside the callable. Otherwise, the callable returns whatever returned by the `callback` delegate.

twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
```diff
@@ -170,11 +172,57 @@ public void testKafkaClient() throws Exception {
...
+    t1.start();
+    t1.join();
+    Thread t2 = createPublishThread(kafkaClient, topic, Compression.NONE, "Testing message", 10, 10);
+    t2.start();
+    t2.join();
+    Thread t3 = createPublishThread(kafkaClient, topic, Compression.SNAPPY, "Snappy Testing message", 10, 20);
+    t3.start();
+    t3.join();
+
+    // 15 messages will be counted since onReceived returns `message.getNextOffset() + 1` as next offset to read
```
chtyim: I don't think the test is correct. You published 30 messages in three message set, hence the `onReceived` method will be called three times. The first time with messages 0-9 and you return 11. The second call with 11-19, and you return 20. The last call with 21-29. So in total there will be more than 15 messages.

twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
```diff
@@ -189,6 +194,49 @@ public void finished() {
...
+      @Override
+      public void finished() {
+        stopLatch.countDown();
+      }
+    });
+    // 15 messages should be in the queue since onReceived returns `message.getNextOffset() + 1` as next offset to read
+    for (int i = 0; i < 30; i += 2) {
+      Assert.assertTrue(i == offsetQueue.poll(60, TimeUnit.SECONDS));
+    }
+    Assert.assertEquals(0, offsetQueue.size());
```
chtyim: should do a `Assert.assertNull(offsetQueue.poll(2, TimeUnit.SECONDS))` instead.

twill-core/src/main/java/org/apache/twill/kafka/client/KafkaConsumer.java
```diff
@@ -33,14 +33,16 @@
 
     /**
      * Invoked when new messages is available.
+     * @param startOffset Offset of the first {@link FetchedMessage} in the iterator of new messages.
```
chtyim: Is this true? If someone specified `-1L` or `-2L` as the offset, would the `startOffset` here be `-1L` or `-2L` or be the actual offset (which won't be negative) of the first message in the iterator?

maochf: How about "The default offset to return as the offset to fetch next message if no message is consumed in this method"

chtyim: or you change the code in a way that this statement is correct.

maochf: If one specified `-1L` or `-2L`, but no message is consumed in the method, then `-1L` or `-2L` still has to be returned

chtyim: Is it? Or shall it return the offset of the first message being read? Other, e.g. if `-1L` was specified, and the latest message was read, but not being consumed by the `onReceived` method. If it returns `-1L`, then that particular message won't get fetched again if there is a newer message available. Is that the intention?

maochf: Yes, I think one should be able to use `-1L` to fetch an upcoming new message, which also aligns with the usage of `-1L` in Kafka's `SimpleConsumer`. If the latest message is read, and it's the first message in the iterator, one can just specify `startOffset` as the latest message offset if one doesn't want to skip it. If the latest message is not the first message in the iterator and its previous message was consumed, `onReceived` will still return the latest message offset.

chtyim: I understand the behavior, but is it a good API? As you see, it is quite difficult to explain, hence documenting it correctly. I feel it's much easier to explain and to use if we say the `startOffset` is the offset of the first message in the given iterator.

maochf: I see. So keep the original description of `startOffset`?

chtyim: The description yes, but I think you need to change the implementation to honor it.

maochf: Yes, the current implementation follows this at line 454. Is just the caller's responsibility to honor this?

chtyim: Not exactly. E.g. let say the offset submitted to Kafka for fetching is "0", but the message with offset "0" is already gone, so the actual first message fetched is having a larger offset.

maochf: how can I get the offset of this first message in this case? `message.getNextOffset()--`?

chtyim: Wrap it with a peeking iterator from guava

maochf: Then `startOffset` is not necessary here? `onReceived` can just keep the first message's offset by itself and the caller should guarantee that no empty iterator is passed to `onReceived` as in `SimpleKafkaConsumer`. Or should we still keep `startOffset` to allow empty iterator to be passed in?

chtyim: I think the `FetchedMessage` doesn't carry the "current" offset, but the next message offset, right? Meaning one can not keep consuming the same message (e.g. there is some failure that it needs to retry on the same message).

chtyim:
LGTM. Please squash the commits to one.

maochf:
Thank you! Just squashed.

PR #17, opened 2016-12-12
Change initial application submission log message from DEBUG to INFO.

anwar6953:
Should be ok to set this log message to INFO, since it only happens once.
In case there is some issue with the yarn app, having this at INFO helps.
There are later logs at INFO, but if the issue causes the code to not reach there, then that isn't sufficient.


yaojiefeng:
LGTM

chtyim:
LGTM.

PR #18, opened 2016-12-13
Fix reset log level for runnables started up later

yaojiefeng:
There is a bug when we try to reset log levels for some runnables which are started up later, e.g, restart or instance changes, after some runtime log level changes. For now, when we start up the `TwillContainerService`, we set log levels based on the what we try to set in `TwillPreparer` + runtime log levels. But we do not memorize the old log level for loggers changed at runtime. So we we reset, we will not get back to the state we want.
This PR fixes this by memorizing the log level changes coming from runtime. And remove when we call reset.


chtyim:
LGTM

PR #19, opened 2016-12-13
Fix ConcurrentModificationException from removal in map and refactor …

yaojiefeng:
…the code
Removing loglevel from the map during iteration will cause ConcurrentModificationException. 
Now Using Iterator to iterate the map and remove. Also refactored the code a bit.


chtyim:
LGTM

PR #20, opened 2016-12-14
Remove unnecessary methods to get the root log level

yaojiefeng:
Since we are now setting the log level once the `TwillContainerService` starts up. We no longer need the get root log level from `TwillContainerMain`.


chtyim:
LGTM

PR #21, opened 2017-01-06
(TWILL-63) Speed up application launch time

chtyim:
The general approach is better jar files management and to cache and reuse jar files created through
class dependency tracing. The changes are further broken down as follows:

1. Refactor jars generation
   - One jar containing the TwillLauncher (launcher.jar), created through dependency tracing.
      - This jar is the same for all applications.
   - One jar containing all twill classes (twill.jar), created through dependency tracing.
      - This jar is the same for all applications.
   - One jar containing the application class, created through dependency tracing.
      - This jar is generated based on the application being launched. It is reusable when launching the same app multiple times.
   - One jar containing user resources setup through TwillPreparer.
      - This jar is not reused between apps.
   - One jar containing runtime config needed by Twill
      - logback.xml, jvm opts, environment, classpaths, ... etc
2. Let YARN to expand jars when localizing to containers instead of expanding it programatically
   - This save time in jar expansion when multiple containers are running on the same host
3. Introduce a new configuration "twill.location.cache.dir" to enable jar caching and reuse
   - Currently only the launcher.jar, twill.jar and application jar will be cached and reuse when possible
   - Cache cleanup logic is also in place to remove files in cache directory that is no longer used by application
4. The ApplicationBundler is improved to allow more flexible usage


twill-yarn/src/main/java/org/apache/twill/yarn/LocationCacheCleaner.java
```diff
@@ -0,0 +1,206 @@
...
+
+    long getExpireTime() {
+      return expireTime;
+    }
+
+    boolean deleteIfExpired(long currentTime) throws IOException {
+      if (currentTime < expireTime) {
+        return false;
+      }
+      location.delete();
```
albertshau: shouldn't this return location.delete()?

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillRunnerService.java
```diff
@@ -277,9 +281,16 @@ public TwillPreparer prepare(TwillApplication application) {
     Preconditions.checkState(serviceDelegate.isRunning(), "Service not start. Please call start() first.");
     final TwillSpecification twillSpec = application.configure();
     final String appName = twillSpec.getName();
+    RunId runId = RunIds.generate();
+    Location appLocation = locationFactory.create(String.format("/%s/%s", twillSpec.getName(), runId.getId()));
+    LocationCache locationCache = this.locationCache;
+    if (locationCache == null) {
+      locationCache =  new NoCachingLocationCache(appLocation);
```
albertshau: nit: extra space

twill-yarn/src/main/java/org/apache/twill/yarn/LocationCacheCleaner.java
chtyim: It is intentional to ignore the result of `delete()` call. When this method returns true, it means it is expired and delete was attempt so that it can be removed from the `pendingCleanups` set. Logically we don't care if the location actually be deleted successfully (if the location is already gone, we don't care. If fail to delete, it will get pick up again in the next cycle).

I'll add a log and also a comment to make this clear.

twill-core/src/main/java/org/apache/twill/internal/ApplicationBundler.java
```diff
@@ -97,10 +76,10 @@ public ApplicationBundler(Iterable<String> excludePackages) {
   }
 
   /**
-   * Constructs a ApplicationBundler.
+   * Constructs an ApplicationBundler.
    *
    * @param excludePackages Class packages to exclude
-   * @param includePackages Class packages that should be included. Anything in this list will override the
+   * @param includePackages Class packages that should be includwwed. Anything in this list will override the
```
chtyim: fixed

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillRunnerService.java
chtyim: fixed

chtyim:
Also rebased on `master` branch to resolve conflicts in `YarnTwillController`.

PR #22, opened 2017-01-07
(TWILL-187) Added max start time

chtyim:


twill-api/src/main/java/org/apache/twill/api/TwillPreparer.java
```diff
@@ -257,8 +258,19 @@
   TwillPreparer setLogLevels(String runnableName, Map<String, LogEntry.Level> logLevelsForRunnable);
 
   /**
-   * Starts the application.
+   * Starts the application. It's the same as calling {@link #start(long, TimeUnit)} with timeout of 60 seconds.
```
gokulavasan: Is there any constant that we can refer to for '60' ?

chtyim:
For the stop timeout, it is cover by TWILL-192, as the changes are more complicate in the stop case

twill-api/src/main/java/org/apache/twill/api/TwillPreparer.java
chtyim: There is, however it's defined in the core module, hence can't refer to it in the API module

PR #23, opened 2017-01-17
(TWILL-181) allow setting the maximum number of retries per runnable

serranom:
feature is complete.  submitting for review.

### DIFF
--- a/twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
+++ b/twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
@@ -37,9 +37,11 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.Service;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+
 import org.apache.hadoop.yarn.api.records.ContainerState;
 import org.apache.twill.api.ResourceReport;
 import org.apache.twill.api.RunId;
+import org.apache.twill.api.RuntimeSpecification;
 import org.apache.twill.api.TwillRunResources;
 import org.apache.twill.api.logging.LogEntry;
 import org.apache.twill.filesystem.Location;
@@ -79,6 +81,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+
 import javax.annotation.Nullable;
 
 /**
@@ -113,9 +116,11 @@ final class RunningContainers {
   private final Location applicationLocation;
   private final Set<String> runnableNames;
   private final Map<String, Map<String, String>> logLevels;
+  private final Map<String, Integer> maxRetries;
+  private final Map<String, Map<Integer, AtomicInteger>> numRetries;
 
   RunningContainers(String appId, TwillRunResources appMasterResources, ZKClient zookeeperClient,
-                    Location applicationLocation, Set<String> runnableNames) {
+    Location applicationLocation, Map<String, RuntimeSpecification> runnables, Map<String, Integer> maxRetriesMap) {
     containers = HashBasedTable.create();
     runnableInstances = Maps.newHashMap();
     completedContainerCount = Maps.newHashMap();
@@ -126,8 +131,10 @@ final class RunningContainers {
     zkClient = zookeeperClient;
     containerStats = HashMultimap.create();
     this.applicationLocation = applicationLocation;
-    this.runnableNames = runnableNames;
+    this.runnableNames = runnables.keySet();
     this.logLevels = new TreeMap<>();
+    this.maxRetries = Maps.newHashMap(maxRetriesMap);
+    this.numRetries = Maps.newHashMap();
   }
 
   /**
@@ -256,6 +263,10 @@ final class RunningContainers {
       if (removeContainerInfo(containerId)) {
         containers.remove(runnableName, containerId);
         removeInstanceId(runnableName, instanceId);
+
+        // clear the entry from numRetries since we are intentionally stopping this instance
+        numRetries.get(runnableName).remove(instanceId);
+
         resourceReport.removeRunnableResources(runnableName, containerId);
         containerChange.signalAll();
       }
@@ -394,6 +405,7 @@ final class RunningContainers {
     try {
       containers.clear();
       runnableInstances.clear();
+      numRetries.clear();
       containerStats.clear();
     } finally {
       containerLock.unlock();
@@ -434,9 +446,12 @@ final class RunningContainers {
       }
 
       boolean containerStopped = false;
+      final String runnableName = lookup.keySet().iterator().next();
+      int instanceId = 0;
+
       for (Map.Entry<String, TwillContainerController> completedEntry : lookup.entrySet()) {
-        String runnableName = completedEntry.getKey();
         TwillContainerController controller = completedEntry.getValue();
+        instanceId = getInstanceId(controller.getRunId());
 
         // TODO: Can there be multiple controllers for a single container?
         // TODO: What is the best way to determine whether to restart container when there are multiple controllers?
@@ -456,13 +471,14 @@ final class RunningContainers {
         removeInstanceId(runnableName, getInstanceId(controller.getRunId()));
         resourceReport.removeRunnableResources(runnableName, containerId);
       }
+      
 
       if (exitStatus != ContainerExitCodes.SUCCESS) {
         LOG.warn("Container {} exited abnormally with state {}, exit code {}.",
                  containerId, state, exitStatus);
-        if (!containerStopped && shouldRetry(exitStatus)) {
+        if (!containerStopped && shouldRetry(runnableName, instanceId, exitStatus)) {
           LOG.info("Re-request the container {} for exit code {}.", containerId, exitStatus);
-          restartRunnables.add(lookup.keySet().iterator().next());
+          restartRunnables.add(runnableName);
         } else if (containerStopped) {
           LOG.info("Container {} is being stopped, will not re-request", containerId);
         }
@@ -477,10 +493,30 @@ final class RunningContainers {
     }
   }
 
-  private boolean shouldRetry(int exitCode) {
-    return exitCode != ContainerExitCodes.SUCCESS
-      && exitCode != ContainerExitCodes.DISKS_FAILED
-      && exitCode != ContainerExitCodes.INIT_FAILED;
+  private boolean shouldRetry(String runnableName, int instanceId, int exitCode) {
+    boolean possiblyRetry = 
+      exitCode != ContainerExitCodes.SUCCESS && 
+      exitCode != ContainerExitCodes.DISKS_FAILED && 
+      exitCode != ContainerExitCodes.INIT_FAILED;
+    
+    if (possiblyRetry) {
+      int max = getMaxRetries(runnableName);
+      if (max == Integer.MAX_VALUE) {
+        return true; // retry without special log msg
+      }
+
+      final int retryCount = getRetryCount(runnableName, instanceId);
+      if (retryCount == max) {
+        LOG.info("Retries exhausted for instance {} of runnable {}.", instanceId, runnableName);
+        return false;
+      } else {
+        LOG.info("Attempting {} of {} retries for instance {} of runnable {}.",
+          retryCount + 1, max, instanceId, runnableName);
+        return true;
+      }
+    } else {
+      return false;
+    }
   }
 
   private boolean isControllerStopped(TwillContainerController controller) {
@@ -525,7 +561,15 @@ final class RunningContainers {
       instances = new BitSet();
       runnableInstances.put(runnableName, instances);
     }
-    int instanceId = instances.nextClearBit(0);
+    
+    // get the next free instance that has not exceeded its maximum retries.
+    int instanceId = 0;
+    int maxRetries = getMaxRetries(runnableName);
+    while (getRetryCount(runnableName, (instanceId = instances.nextClearBit(instanceId))) == maxRetries) {
+      instanceId++;
+    }
+    // count starts at -1, so first increment makes it 0
+    incrementRetryCount(runnableName, instanceId);
     instances.set(instanceId);
     return instanceId;
   }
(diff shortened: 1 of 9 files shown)


hsaputra:
Thanks for the PR. Could you kindly add more in the PR description on how the solution is implemented.
This will help review to "compare" the code implementation with the description of what you were trying to do.

serranom:
@hsaputra , it follows the design in the ticket.  should those details be repeated here?

serranom:
Also, I noticed the test failure and I'm confused about it.  It seems to be something intermittent because it succeeds in some cases.  When it fails, the AM doesn't restart the containers:

    2017-01-17T23:43:19,872Z WARN  o.a.t.i.a.RunningContainers [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] RunningContainers:handleCompleted(RunningContainers.java:476) - Container container_1484695869902_0027_01_000006 exited abnormally with state COMPLETE, exit code 1.
    2017-01-17T23:43:19,872Z INFO  o.a.t.i.a.RunningContainers [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] RunningContainers:shouldRetry(RunningContainers.java:518) - 5 of 6 retries for runnable FailingServer.
    2017-01-17T23:43:19,873Z INFO  o.a.t.i.a.RunningContainers [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] RunningContainers:handleCompleted(RunningContainers.java:479) - Re-request the container container_1484695869902_0027_01_000006 for exit code 1.
    2017-01-17T23:43:19,873Z INFO  o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:handleCompleted(ApplicationMasterService.java:502) - Re-request container for FailingServer with 1 instances.
    2017-01-17T23:43:19,873Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:19,873Z INFO  o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:444) - Relaxing provisioning constraints for request f30d39fe-2d65-493d-b670-8b42192bf322
    2017-01-17T23:43:20,875Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:21,878Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:22,880Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:23,882Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:24,884Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:24,885Z INFO  o.a.t.i.LogOnlyEventHandler [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] LogOnlyEventHandler:launchTimeout(LogOnlyEventHandler.java:36) - Requested 2 containers for runnable FailingServer, only got 0 after 5012 ms.
    2017-01-17T23:43:25,888Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:26,890Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3
    2017-01-17T23:43:27,892Z DEBUG o.a.t.i.a.ApplicationMasterService [testing-docker-7dddd7e0-951d-4746-842a-bd7d0e7c25ce] [ApplicationMasterService] ApplicationMasterService:doRun(ApplicationMasterService.java:406) - Runnable container requests: 3

Any advice on debugging this?

hsaputra:
Ah, I see you out the proposed design as JIRA comment.
Usually the design in JIRA changes but once you start implementing, that's why I usually ask for developers to add the actual design or plan implemented in the PR

serranom:
Gotcha.  Here it is, cleaned up for what I actually did:

* Each runnable can have a configured number of max retries.  If not set, then retries are unlimited as before.
* add withMaxTries(runnableName, int) to TwillPreparer
* add withMaxTries(runnableName, int) to YarnTwillPreparer.  This stores a map from runnableName to maxRetries.
* this map becomes part of the twillRuntimeSpecification and RuntimeSpecification interface and is added to TwillRuntimeSpecificationCodec
* ApplicationMasterService.initRunningContainers is updated to pass a map of runnables to maxretries.
* updated RunningContainers so that it keeps count of the number of retries per runnable and uses this in handleCompleted() to determine if it should retry. Since every instance is the same as any other, if I'm starting 10 instances of a Runnable, and wanted a max retry count of 3, then that would scale the total number of retries to 30. Each instance gets (on average) 3 tries. Since the instances are interchangeable, there is no concept of a discrete instance being retried.
* updated logging to not have anything special if max wasn't set and to log the number of retries left and when they have been exhausted.

serranom:
A couple of questions.  Note, I'm not irritated or anything like that... just want to have my expectations in line for planning purposes.

* What should I generally expect for feedback on the merging of a pull request?
* Is there any issue with this pull request (code-wise) that contributes to the time?
* Does the failing CI check (which I've seen is intermittent) a contributor to the time?  That is the fact that it has a failure means it would get less attention
* Should I be doing anything different?

Thanks!

hsaputra:
Hi @serranom , usually in about week you should get review. I apologize for the delay, but sometimes there is delay just due to volunteer time.
As for CI failures, it depends, sometime ZK is flaky so functional tests could fail.

You are doing right thing, we as committers needs to update our level of service. Thanks for your patience

hsaputra:
Picking up this PR for review

hsaputra:
Let's trigger new build, @serranom could you close and reopen this PR please?

hsaputra:
Reference: https://issues.apache.org/jira/browse/TWILL-181

Hi @serranom , Could you also update the PR title to prefix with `(TWILL-181)` as shown in how to contribute page [1]?
This will trigger the Git hook to close the JIRA when PR is merged.

[1] http://twill.apache.org/HowToContribute.html

serranom:
Thanks Henry!  No rush on my account, I was just checking in.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -113,9 +117,11 @@ public Integer apply(BitSet input) {
   private final Location applicationLocation;
   private final Set<String> runnableNames;
   private final Map<String, Map<String, String>> logLevels;
+  private final Map<String, Integer> maxRetries;
```
poornachandra: The max retries count in this map will have to be updated when the instance count of a runnable changes, right?

serranom: Good catch.  Yes.  I missed that.  I will update the pull request with a fix.  When the number of instances increases, `deltaInstances * maxRetries` will be added to this instance adjusted maximum.  When the number of instances decreases, the `deltaInstance*maxRetries` will be subtracted from the instance adjusted maximum.  Additionally upon a decrease, if the number of retries exceeds the instance adjusted maximum it will be set to the instance adjusted maximum.

serranom:
@hsaputra, since I have a bug to fix, the updated request will trigger the rebuild.  thanks.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
poornachandra: On further thought, I think if we track the restarts per instance id then it would simplify the contract for max retries. This will be consistent with what the javadoc says too -
```
Sets the maximum number of times (per instance) a runnable will be retried if it exits without success. The default behavior is to retry indefinitely.
```

serranom: from my analysis the association of instance ids does not necessarily correspond to specific processes.  it looked like there is a pool of requests, a new request being serviced gets the lowest instance id and a failed request gets put back on the queue.  this is why i went with an instance adjusted number of retries.  did i miss something?

poornachandra: Instance ids for runnable instances go from `0` to `num-instances - 1`. If an instance gets restarted, then it retains the same instance id. Using this instance id we can track the number of times a particular instance was restarted.

You can get the instance id in `handleCompleted` method by using `getInstanceId(RunId runId)` method.

hsaputra:
Seems like one of the tests failed related to new one you added: ` MaxRetriesTestRun.maxRetriesTwoInstances`

Could you quickly take a peek?
https://s3.amazonaws.com/archive.travis-ci.org/jobs/192819935/log.txt

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
serranom: I'm taking a look at this.

serranom:
@hsaputra , that is the failure I see as intermittent.  My comment above has a relevant log excerpt.  I'm planning on trying to dig into it today or tomorrow.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
serranom: @poornachandra, I don't see anything linking one instance request to a retry of that failure.  Failures get added to the multiset which then results in a new request for that many instances.  This means in the end, there is nothing tying one attempt for a particular instanceId to the next one.  So we could keep track of retries by instanceId, but it would end up functionally being the same as keeping track of retries for the entire Runnable.  I think it would just add more complexity for little benefit to track by instanceId.

serranom: but on second thought, the additional complexity of tracking by instance id would handle the following case better:

- initially start with `x` instances, all of which succeed.
- add 1 more instance, which fails

If tracking by instance id, the new instance will get maxRetries.  Without tracking by instance id, the new instance would get `maxRetries*(x+1)` retries.  So I will add the complexity of keeping track by id.  It seems worth it.

poornachandra: 👍

serranom:
I have discovered the cause of the intermittent failure.

This code (starting on line 703 of ApplicationMasterService, comments added by me):

      /*
       * The provisionRequest will either contain a single container (ALLOCATE_ONE_INSTANCE_AT_A_TIME), or all the
       * containers to satisfy the expectedContainers count. In the later case, the provision request is complete once
       * all the containers have run at which point we poll() to remove the provisioning request.
       */
      if (expectedContainers.getExpected(runnableName) == runningContainers.count(runnableName) ||
        provisioning.peek().getType().equals(AllocationSpecification.Type.ALLOCATE_ONE_INSTANCE_AT_A_TIME)) {
        provisioning.poll();
      }

There is a case when instances are failing (but not simultaneously) where the retries for the instances will be spread over two invocations of `ApplicationMasterService.handleCompleted`.  This means they will be part of separate `RunnableContainerRequests` and thus will be provisioned separately.  But because the code above does not anticipate this case, the first provisionRequest will never appear to be satisfied, never be polled and the total can never be met.  This is an existing bug, since I think it could happen before my new code -- its just that my test tickles it.  If there is agreement, I can file a new JIRA and PR request or just fix it with this PR.   Let me know.

serranom:
I have new changes I will add to this PR once https://github.com/apache/twill/pull/29 is resolved since the tests will fail intermittently without that fix.

serranom:
I think this is set now.  I've updated the PR with instance-based maxRetries changes.

serranom:
ugh.  I'll take a look, I may have messed up the merge.

serranom:
I found that there is still a race condition in ApplicationMasterService.launchRunnable.  If the number of instances is increased right after the original request is fullfilled, the current logic can result in the original request not being polled resulting in future requests hanging.  This seems to be a fairly unlikely case in the real world, but I'll file a JIRA for it.  For now, I've reworked the test to wait until launchRunnable has done the polling for the original request.  This PR is now complete.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -477,10 +493,30 @@ void handleCompleted(YarnContainerStatus status, Multiset<String> restartRunnabl
...
+        exitCode != ContainerExitCodes.DISKS_FAILED && 
+        exitCode != ContainerExitCodes.INIT_FAILED;
+    
+    if (possiblyRetry) {
+      int max = getMaxRetries(runnableName);
+      if (max == Integer.MAX_VALUE) {
+        return true; // retry without special log msg
+      }
+
+      if (getRetryCount(runnableName, instanceId) == max) {
```
hsaputra: Since we call `getRetryCount` for this check, might as well cache it as local var to make sure we got  the right one for this request:

```
int retryCount = getRetryCount(runnableName, instanceId);
if (getRetryCount(runnableName, instanceId) == max) {
  ...
} else {
  LOG.info("Attempting {} of {} retries for instance {} of runnable {}.", retryCount + 1,
    max, instanceId, runnableName);
  return true;
}
```

hsaputra:
Thanks much for working on the patch and for the update, @serranom. 

Did a pass for review.
LGTM in general, add mini comment. Other than that +1 will merge after comment addressed.

serranom:
My pleasure.  It was a good way to get to know twill code base.

hsaputra:
I am having problem applying the patch, could you kindly rebase the patch into single commit?
Thanks

serranom:
I looked at the latest CI run and the failure is related to some `twill-zookeeper` project issue -- not yarn

hsaputra:
Looks like not related to this PR. Will merge EOD if no more comment.

Thanks again for the hard work, @serranom

PR #24, opened 2017-01-19
Twill 90

Esquive:
Hello, 

for a project I required the capability to set the resources of the YARN application Manager. In the solution I propose I did not break the API all the Builders are extended to support the following AM resources: MEM and CPU cores. I extended the TwillSpecification classes to accept the said parameters. I did not create a dedicated "TwillAMSpecification" subclass as AM is also a "TwillRunnable".

Cheers


chtyim:
Please use 2 spaces instead of 4 for indentation

PR #26, opened 2017-01-24
(TWILL-208) add Location.mkdirs(String permissions)

anew:
- adds mkdirs() that takes permissions
- makes sure getOutputStream(permisssions) also uses that when creating the directories 
- add test case

### DIFF
--- a/twill-common/src/main/java/org/apache/twill/filesystem/LocalLocation.java
+++ b/twill-common/src/main/java/org/apache/twill/filesystem/LocalLocation.java
@@ -98,8 +98,8 @@ final class LocalLocation implements Location {
 
   @Override
   public OutputStream getOutputStream(String permission) throws IOException {
-    ensureDirectory(file.getParentFile());
     Set<PosixFilePermission> permissions = parsePermissions(permission);
+    ensureDirectory(file.getParentFile(), permissions);
     Path path = file.toPath();
     WritableByteChannel channel = Files.newByteChannel(path,
                                                        EnumSet.of(StandardOpenOption.CREATE, StandardOpenOption.WRITE),
@@ -110,7 +110,7 @@ final class LocalLocation implements Location {
   }
 
   /**
-   * @return Returns the name of the file or directory denoteed by this abstract pathname.
+   * @return Returns the name of the file or directory denoted by this abstract pathname.
    */
   @Override
   public String getName() {
@@ -251,6 +251,34 @@ final class LocalLocation implements Location {
     return file.mkdirs();
   }
 
+  @Override
+  public boolean mkdirs(String permission) throws IOException {
+    Set<PosixFilePermission> posixPermissions = parsePermissions(permission);
+    return mkdirs(file, posixPermissions);
+  }
+
+  private boolean mkdirs(File file, Set<PosixFilePermission> permissions) throws IOException {
+    if (file.exists()) {
+      return false;
+    }
+    if (file.mkdir()) {
+      Files.setPosixFilePermissions(file.toPath(), permissions);
+      return true;
+    }
+    File parent = file.getParentFile();
+    if (parent == null) {
+      return false;
+    }
+    if (!mkdirs(parent, permissions) && !parent.exists()) {
+      return false;
+    }
+    if (file.mkdir()) {
+      Files.setPosixFilePermissions(file.toPath(), permissions);
+      return true;
+    }
+    return false;
+  }
+
   /**
    * @return Length of file.
    */
@@ -317,6 +345,16 @@ final class LocalLocation implements Location {
   private void ensureDirectory(File dir) throws IOException {
     // Check, create, check to resolve race conditions if there are concurrent creation of the directory.
     if (!dir.isDirectory() && !dir.mkdirs() && !dir.isDirectory()) {
+      throw new IOException("Failed to create directory " + dir);
+    }
+  }
+
+  /**
+   * Ensures the given {@link File} is a directory. If it doesn't exist, it will be created.
+   */
+  private void ensureDirectory(File dir, Set<PosixFilePermission> permission) throws IOException {
+    // Check, create, check to resolve race conditions if there are concurrent creation of the directory.
+    if (!dir.isDirectory() && !mkdirs(dir, permission) && !dir.isDirectory()) {
       throw new IOException("Failed to create directory " + dir);
     }
   }
--- a/twill-common/src/main/java/org/apache/twill/filesystem/Location.java
+++ b/twill-common/src/main/java/org/apache/twill/filesystem/Location.java
@@ -173,7 +173,7 @@ public interface Location {
    * pathname denotes a directory, then the directory must be empty in order
    * to be deleted.
    *
-   * @return true if and only if the file or directory is successfully delete; false otherwise.
+   * @return true if and only if the file or directory is successfully deleted; false otherwise.
    */
   boolean delete() throws IOException;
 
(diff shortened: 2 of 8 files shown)


twill-common/src/main/java/org/apache/twill/filesystem/Location.java
```diff
@@ -207,6 +207,21 @@
...
+   * but nonexistent parent directories.
+   *
+   * @param permission A permission string. It has to be either a three digit or a nine character string.
+   *                   For the three digit string, it is similar to the UNIX permission numeric representation.
+   *                   The first digit is the permission for owner, second digit is the permission for group and
+   *                   the third digit is the permission for all.
+   *                   For the nine character string, it uses the format as specified by the
+   *                   {@link PosixFilePermissions#fromString(String)} method.
+   *
+   * @return true if and only if the renaming succeeded; false otherwise
```
albertshau: update javadoc

poornachandra: `@return true if and only if the renaming succeeded; false otherwise` => needs to change from renaming to mkdir

twill-common/src/main/java/org/apache/twill/filesystem/LocalLocation.java
```diff
@@ -98,8 +98,8 @@ public OutputStream getOutputStream() throws IOException {
 
   @Override
   public OutputStream getOutputStream(String permission) throws IOException {
-    ensureDirectory(file.getParentFile());
     Set<PosixFilePermission> permissions = parsePermissions(permission);
+    ensureDirectory(file.getParentFile(), permissions);
```
poornachandra: Do we want to use the file permissions for the directory too?

twill-common/src/main/java/org/apache/twill/filesystem/LocalLocation.java
```diff
@@ -251,6 +251,34 @@ public boolean mkdirs() throws IOException {
     return file.mkdirs();
   }
 
+  @Override
+  public boolean mkdirs(String permission) throws IOException {
```
poornachandra: Should we use  [Files.createDirectories](http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createDirectories(java.nio.file.Path,%20java.nio.file.attribute.FileAttribute...)) since it also takes permissions?

twill-common/src/main/java/org/apache/twill/filesystem/Location.java
anew: good catch. I actually did not want that line :)

twill-common/src/main/java/org/apache/twill/filesystem/LocalLocation.java
anew: yes, the intention of permissions is that others get access through, for example, group permission. So if I create the file with 700 but the directories are created with the default umask of 750, then that is not great. Even worse, if I create the file with 755, and the directories are created with 750, then world cannot access the file despite the intention.

twill-common/src/main/java/org/apache/twill/filesystem/LocalLocation.java
anew: oh, did not know about that. let me try.

anew: hmm that does not seem to apply the permissions correctly, it applies the umask to the permissions. What we want is the exact permissions as passed in.

twill-yarn/src/test/java/org/apache/twill/filesystem/HDFSLocationTest.java
```diff
@@ -47,4 +52,14 @@ public static void finish() {
   protected LocationFactory createLocationFactory(String pathBase) throws Exception {
     return new HDFSLocationFactory(dfsCluster.getFileSystem(), pathBase);
   }
+
+  // TODO (TWILL-209): figure out how to make MiniDFSCluster enforce permissions
+  @Ignore
+  @Test(expected = Exception.class)
+  public void testPermissions() throws IOException {
+    // create a directory that does not permit anything
+    dfsCluster.getFileSystem().mkdir(new Path("/a"), FsPermission.valueOf("----------"));
```
gokulavasan: How do we know for sure that the exception was not thrown by this operation?

twill-common/src/main/java/org/apache/twill/filesystem/Location.java
anew: done

twill-yarn/src/test/java/org/apache/twill/filesystem/HDFSLocationTest.java
anew: Good point. Anyway, this test does not throw an exception and is ignored because of that. I guess this is a typical problem with using expected=, we can never know whether it is thrown by the line in the code where we expect. So should we rewrite all our test cases not to use that? And explicitly fail if the exception is not thrown?

gokulavasan: Oh ok, missed the Ignore annotation. It might be good to try and catch and have an Assert.fail at the end of the try block for the new tests that we are adding.

anew: updated.

twill-yarn/src/test/java/org/apache/twill/filesystem/HDFSLocationTest.java
```diff
@@ -55,11 +55,19 @@ protected LocationFactory createLocationFactory(String pathBase) throws Exceptio
...
-    dfsCluster.getFileSystem().mkdir(new Path("/a/b"), FsPermission.valueOf("----------"));
+    boolean succeeded = false;
+    try {
+      // creating a subdir should fail because even the owner has no write permission
+      dfsCluster.getFileSystem().mkdir(new Path("/a/b"), FsPermission.valueOf("----------"));
+      succeeded = true;
+    } catch (Exception e) {
+      // expected; TODO: figure out which exception should the expected here, it's not documented
+    }
+    if (succeeded) {
```
gokulavasan: nit: you can move the Assert.fail to line 65 and remove the boolean variable.

anew: no I can't because line 66 would catch it.

gokulavasan: No it won't I think. Since Assert.fail throws AssertionError which doesn't extend Exception?

anew: oh right

gokulavasan:
LGTM 👍

twill-common/src/main/java/org/apache/twill/filesystem/LocalLocation.java
poornachandra: I see, in that case I think it would be good to add this to the javadoc of the method.

poornachandra:
Just one comment on javadoc, please fix and merge 👍

poornachandra:
LGTM

anew:
Squashing commits now.

twill-yarn/src/test/java/org/apache/twill/filesystem/FileContextLocationTest.java
```diff
@@ -17,26 +17,36 @@
...
   public static MiniDFSCluster dfsCluster;
+  private static UserGroupInformation testUGI;
 
   @BeforeClass
   public static void init() throws IOException {
     Configuration conf = new Configuration();
-    conf.setBoolean("hdfs.permissions", true);
     conf.set(MiniDFSCluster.HDFS_MINIDFS_BASEDIR, tmpFolder.newFolder().getAbsolutePath());
     dfsCluster = new MiniDFSCluster.Builder(conf).numDataNodes(1).build();
+    // make rooot world-writable so that we can create all location factories as unprivileged user
```
chtyim: root

poornachandra:
LGTM

anew:
squashing one more time

PR #27, opened 2017-01-27
Fix location permission test for older Hadoop versions

anew:
In older Hadoop versions (2.0), files are always created with permissions that do not have execute bits. In later Hadoop versions, execute permissions are permitted for files. This changes the test location permission case to adjust the expected permissions for the missing execute bits, based on the Hadoop version.  

Also fixes an intriguing bug where creating a directory (say a/b/c/) one of whose parents (say a/b) already exists as a file, Hadoop throws an access control exception. This is because the execute bit is not set on files. We must expect that exception and handle it properly.


chtyim:
LGTM

PR #29, opened 2017-01-29
(TWILL-211) use ALLOCATE_ONE_INSTANCE_AT_A_TIME for retries to prevent…

serranom:
… poll starvation.

These changes fix a lingering bug identified by the tests for TWILL-181.  When creating a runnable container request, we check for number of instances equal to one.  For such cases we always set the allocation request to use `ALLOCATE_ONE_INSTANCE_AT_A_TIME`.  When retries occur, each failed instance results in a single provision request, each asking for one instance.  But if `ALLOCATE_ONE_INSTANCE_AT_A_TIME` is _not_ set, the `handleCompleted` code will never consider the request satisfied since it is waiting for all instances to have been started.

The check for `null` at line 469 is necessary because these new provision requests do not have a placement policy.


hsaputra:
This LGTM

+1

hsaputra:
Will merge this if no more comment

poornachandra:
LGTM from me too

PR #31, opened 2017-02-02
Avoid packaging resources jar in the ApplicationMaster

anwar6953:
Avoid packaging resources jar in the ApplicationMaster, so that the logback.xml in it doesn't affect the AM.
Otherwise, the stdout of the AM omits a lot of logs.


chtyim:
LGTM

PR #32, opened 2017-02-13
(TWILL-218), add the OptionSpec class as an explicit dependency for the…

serranom:
… appmaster

I added the jar to the respective poms and added OptionSpec to the appmaster dependency calculation.  The dependency shows up because of Kafka:

    [INFO] |  |  \- org.apache.kafka:kafka_2.10:jar:0.8.0:compile
    [INFO] |  |     +- org.scala-lang:scala-library:jar:2.10.1:compile
    [INFO] |  |     +- net.sf.jopt-simple:jopt-simple:jar:3.2:compile
    [INFO] |  |     +- org.scala-lang:scala-compiler:jar:2.10.1:compile
    [INFO] |  |     |  \- org.scala-lang:scala-reflect:jar:2.10.1:compile
    [INFO] |  |     +- com.101tec:zkclient:jar:0.3:compile
    [INFO] |  |     +- com.yammer.metrics:metrics-core:jar:2.2.0:compile
    [INFO] |  |     \- com.yammer.metrics:metrics-annotation:jar:2.2.0:compile

However the twill dependency mechanism does not detect the dependency.  By making it explicit, it will always end up in the `twill.jar`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -161,6 +161,7 @@
         <kafka.version>0.8.0</kafka.version>
         <zookeeper.version>3.4.5</zookeeper.version>
         <junit.version>4.11</junit.version>
+        <jopt-simple.version>3.2</jopt-simple.version>
         <commons-compress.version>1.5</commons-compress.version>
         <hadoop.version>[2.0.2-alpha,2.3.0]</hadoop.version>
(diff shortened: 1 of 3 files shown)


chtyim:
LGTM

PR #33, opened 2017-02-16
TWILL-216 Make ratio between total memory and on-heap memory configur…

yufeldman:
…able

### DIFF
--- a/twill-api/src/main/java/org/apache/twill/api/Configs.java
+++ b/twill-api/src/main/java/org/apache/twill/api/Configs.java
@@ -34,6 +34,11 @@ public final class Configs {
     public static final String JAVA_RESERVED_MEMORY_MB = "twill.java.reserved.memory.mb";
 
     /**
+     * Configurable ratio between Heap and Reserved Memory
+     */
+    public static final String HEAP_RESERVED_MIN_RATIO = "twill.java.heap.memory.ratio";
+
+    /**
      * Set this to false to disable the secure store updates done by default.
      */
     public static final String SECURE_STORE_UPDATE_LOCATION_ENABLED = "twill.secure.store.update.location.enabled";
@@ -75,6 +80,7 @@ public final class Configs {
      */
     public static final int JAVA_RESERVED_MEMORY_MB = 200;
 
+    public static final double HEAP_RESERVED_MIN_RATIO = 0.7d;
     /**
      * Default use the system temp directory for local staging files.
      */
--- a/twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
@@ -59,12 +59,13 @@ public final class TwillContainerLauncher {
   private final int instanceCount;
   private final JvmOptions jvmOpts;
   private final int reservedMemory;
+  private final double minHeapRatio;
   private final Location secureStoreLocation;
 
   public TwillContainerLauncher(RuntimeSpecification runtimeSpec, ContainerInfo containerInfo,
                                 ProcessLauncher.PrepareLaunchContext launchContext,
                                 ZKClient zkClient, int instanceCount, JvmOptions jvmOpts, int reservedMemory,
-                                Location secureStoreLocation) {
+                                Location secureStoreLocation, double minHeapRatio) {
     this.runtimeSpec = runtimeSpec;
     this.containerInfo = containerInfo;
     this.launchContext = launchContext;
@@ -72,6 +73,7 @@ public final class TwillContainerLauncher {
     this.instanceCount = instanceCount;
     this.jvmOpts = jvmOpts;
     this.reservedMemory = reservedMemory;
+    this.minHeapRatio = minHeapRatio;
     this.secureStoreLocation = secureStoreLocation;
   }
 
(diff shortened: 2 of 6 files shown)


hsaputra:
@yufeldman : Could you kindly add in the summary of the PR the summary of the approach you are taking to fix the issue?

Will help with the review to compare expectation vs implementation.

twill-api/src/main/java/org/apache/twill/api/Configs.java
```diff
@@ -75,6 +82,7 @@ private Keys() {
      */
     public static final int JAVA_RESERVED_MEMORY_MB = 200;
 
+    public static final double HEAP_RESERVED_MIN_RATIO_DEFAULT = Constants.HEAP_MIN_RATIO;
```
chtyim: Inside the `Defaults` class, no need to repeat the variable with `DEFAULT`. Just name it the same as the config key constant.

twill-api/src/main/java/org/apache/twill/api/Configs.java
```diff
@@ -34,6 +36,11 @@
     public static final String JAVA_RESERVED_MEMORY_MB = "twill.java.reserved.memory.mb";
 
     /**
+     * Configurable ratio between Heap and Reserved Memory
+     */
+    public static final String HEAP_RESERVED_MIN_RATIO_CONFIG = "twill.java.reserved.memory.ratio";
```
chtyim: No need to have `_CONFIG` as it is already defined inside the `Keys` inner class.

chtyim: Also, I think the config key should be `twill.java.min.heap.memory.ratio`, as it is the minimum ratio of heap memory.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
```diff
@@ -59,19 +59,21 @@
...
   private final JvmOptions jvmOpts;
   private final int reservedMemory;
+  private final Double heapToReservedRatio;
   private final Location secureStoreLocation;
 
   public TwillContainerLauncher(RuntimeSpecification runtimeSpec, ContainerInfo containerInfo,
                                 ProcessLauncher.PrepareLaunchContext launchContext,
                                 ZKClient zkClient, int instanceCount, JvmOptions jvmOpts, int reservedMemory,
-                                Location secureStoreLocation) {
+                                Location secureStoreLocation, Double heapToReservedRatio) {
```
chtyim: Better to have `minHeapRatio` a `double` and the caller need to make sure it is not null.

twill-api/src/main/java/org/apache/twill/api/Configs.java
chtyim: You can remove the `Constants.HEAP_MIN_RATIO` and just define the default here.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -41,11 +41,23 @@
   private final String rmSchedulerAddr;
   private final Map<String, Map<String, String>> logLevels;
   private final Map<String, Integer> maxRetries;
+  private Double heapToReserveRatio;
```
chtyim: Shouldn't allow null. This is an internal class, you can simply require the caller to pass in valid value.

twill-core/src/main/java/org/apache/twill/internal/json/TwillRuntimeSpecificationCodec.java
```diff
@@ -45,6 +45,7 @@
   private static final String TWILL_RUNID = "twillRunId";
   private static final String TWILL_APP_NAME = "twillAppName";
   private static final String RESERVED_MEMORY = "reservedMemory";
+  private static final String HEAP_TO_RESERVED_RATIO = "heapToReservedRatio";
```
chtyim: for consistency, just have everywhere call it `minHeapRatio`

chtyim:
The approach seems fine. Just couple comments about the naming and not to allow `null` ratio.

twill-core/src/main/java/org/apache/twill/internal/TwillContainerLauncher.java
hsaputra: do we need double here? Would float be ok?

twill-core/src/main/java/org/apache/twill/internal/json/TwillRuntimeSpecificationCodec.java
```diff
@@ -94,6 +96,8 @@ public TwillRuntimeSpecification deserialize(JsonElement json, Type typeOfT,
                                          jsonObj.has(RM_SCHEDULER_ADDR) ?
                                          jsonObj.get(RM_SCHEDULER_ADDR).getAsString() : null,
                                          logLevels,
-                                         maxRetries);
+                                         maxRetries,
+                                         jsonObj.has(HEAP_RESERVED_MIN_RATIO) ?
```
chtyim: Should use HEAP_RESERVED_MIN_RATIO as the default instead of `null`

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -17,6 +17,8 @@
  */
 package org.apache.twill.yarn;
 
+import joptsimple.OptionSpec;
```
chtyim: Please don't reorder import.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -38,7 +38,6 @@
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
```
chtyim: Please don't update import spaces. If you use IntelliJ or eclipse, you can download the style from the twill site.

twill-core/src/main/java/org/apache/twill/internal/json/TwillRuntimeSpecificationCodec.java
yufeldman: sure

chtyim:
LGTM. Will merge it when build pass.

hsaputra:
@chtyim I was asking concern about the selection of double vs float

hsaputra:
Ah, Yulia just replied on JIRA that we HAD ALREADY used double for the ratio (not sure why) but this PR just continuing the existing data format.

+1 then

PR #37, opened 2017-03-17
(Twill-222) fix for default root log level

yaojiefeng:
JIRA: https://issues.apache.org/jira/browse/TWILL-222

We should not assume the root log level as INFO in TwillPreparer. Removed the INFO level from TwillPreparer to not override what is in the logback.xml.


chtyim:
LGTM

PR #39, opened 2017-03-20
(TWILL-225) Refactor TwillPreprer to allow configurations affecting a…

chtyim:
…pplication behavior overridable per TwillPreprer

### DIFF
--- a/twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
+++ b/twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
@@ -40,6 +40,7 @@ import com.google.common.reflect.TypeToken;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import joptsimple.OptionSpec;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.Credentials;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.token.Token;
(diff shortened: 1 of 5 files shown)


twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -155,25 +153,16 @@ public String apply(Class<?> cls) {
   private ClassAcceptor classAcceptor;
   private final Map<String, Integer> maxRetries = Maps.newHashMap();
 
-  YarnTwillPreparer(YarnConfiguration yarnConfig, TwillSpecification twillSpec, RunId runId,
-                    YarnAppClient yarnAppClient, String zkConnectString, Location appLocation, Set<URL> twillClassPaths,
+  YarnTwillPreparer(Configuration config, TwillSpecification twillSpec, RunId runId,
```
yufeldman: Any particular reason to switch to Configuration from YarnConfiguration?

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -455,21 +475,40 @@ private void saveLogLevels(String runnableName, Map<String, LogEntry.Level> logL
...
 
+  /**
+   * Creates an {@link Credentials} by copying the {@link Credentials} of the current user.
+   */
   private Credentials createCredentials() {
     Credentials credentials = new Credentials();
 
     try {
       credentials.addAll(UserGroupInformation.getCurrentUser().getCredentials());
+    } catch (IOException e) {
```
anwar6953: shouldn't this be fatal?

chtyim: I don't think so. Even if this call fail, if the caller is able to acquire required credentials in his own way and add them via the `addSecureStore` method, then the app can still be submitted and executed correctly.

anwar6953:
LGTM

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
anwar6953: I guess app submission would fail if caller doesn't acquire required credentials in its own way anyways.

PR #40, opened 2017-03-21
(TWILL-122) Allow disabling log collection

chtyim:
- Introduced a new configuration twill.log.collection.enabled for
  turning off log collection
- Refactor YarnTwillController and related class hierarchy to not
  starting Kafka client when log collection is disabled
- Added Kafka zk connection string information in AM live node data
- Refactor KafkaAppender and ServiceMain configureLogger
  - Log to StatusManager instead of Logger to avoid recursive logging
  - Instead of resetting logback configuration, directly instantiate and
    add the Kafka log appender to the logging context.
- Refactor ServiceMain, ApplicationMasterMain and TwillContainerMain to
  simplify ZK Connection string construction

### DIFF
--- a/twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
@@ -71,21 +71,37 @@ public abstract class AbstractTwillController extends AbstractZKServiceControlle
   private static final Logger LOG = LoggerFactory.getLogger(AbstractTwillController.class);
   private static final Gson GSON = new Gson();
 
+  private final String appName;
+  private final RunId runId;
   private final Queue<LogHandler> logHandlers;
   private final KafkaClientService kafkaClient;
   private ZKDiscoveryService discoveryServiceClient;
   private Cancellable logCancellable;
 
-  public AbstractTwillController(RunId runId, ZKClient zkClient, Iterable<LogHandler> logHandlers) {
+  public AbstractTwillController(String appName, RunId runId, ZKClient zkClient, boolean logCollectionEnabled,
+                                 Iterable<LogHandler> logHandlers) {
     super(runId, zkClient);
+    this.appName = appName;
+    this.runId = runId;
     this.logHandlers = new ConcurrentLinkedQueue<>();
-    this.kafkaClient = new ZKKafkaClientService(ZKClients.namespace(zkClient, "/" + runId.getId() + "/kafka"));
-    Iterables.addAll(this.logHandlers, logHandlers);
+
+    // When addressing TWILL-147, need to check if the given ZKClient is
+    // actually used by the Kafka used for log collection
+    if (logCollectionEnabled) {
+      this.kafkaClient = new ZKKafkaClientService(ZKClients.namespace(zkClient, "/" + runId.getId() + "/kafka"));
+      Iterables.addAll(this.logHandlers, logHandlers);
+    } else {
+      this.kafkaClient = null;
+      if (!Iterables.isEmpty(logHandlers)) {
+        LOG.warn("Log collection is disabled for application {} with runId {}. " +
+                   "Adding log handler won't get any logs.", appName, runId);
+      }
+    }
   }
 
   @Override
   protected synchronized void doStartUp() {
-    if (!logHandlers.isEmpty()) {
+    if (kafkaClient != null && !logHandlers.isEmpty()) {
       kafkaClient.startAndWait();
       logCancellable = kafkaClient.getConsumer().prepare()
                                   .addFromBeginning(Constants.LOG_TOPIC, 0)
@@ -101,12 +117,20 @@ public abstract class AbstractTwillController extends AbstractZKServiceControlle
     if (discoveryServiceClient != null) {
       discoveryServiceClient.close();
     }
-    // Safe to call stop no matter when state the KafkaClientService is in.
-    kafkaClient.stopAndWait();
+    if (kafkaClient != null) {
+      // Safe to call stop no matter what state the KafkaClientService is in.
+      kafkaClient.stopAndWait();
+    }
   }
 
   @Override
   public final synchronized void addLogHandler(LogHandler handler) {
+    if (kafkaClient == null) {
+      LOG.warn("Log collection is disabled for application {} with runId {}. " +
+                 "Adding log handler won't get any logs.", appName, runId);
+      return;
+    }
+
     logHandlers.add(handler);
     if (logHandlers.size() == 1) {
       kafkaClient.startAndWait();
--- a/twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
+++ b/twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
@@ -41,13 +41,14 @@ public class TwillRuntimeSpecification {
   private final String rmSchedulerAddr;
   private final Map<String, Map<String, String>> logLevels;
   private final Map<String, Integer> maxRetries;
-  private double minHeapRatio;
+  private final double minHeapRatio;
+  private final boolean logCollectionEnabled;
 
   public TwillRuntimeSpecification(TwillSpecification twillSpecification, String fsUser, URI twillAppDir,
                                    String zkConnectStr, RunId twillRunId, String twillAppName,
                                    int reservedMemory, @Nullable String rmSchedulerAddr,
                                    Map<String, Map<String, String>> logLevels, Map<String, Integer> maxRetries,
-                                   double minHeapRatio) {
+                                   double minHeapRatio, boolean logCollectionEnabled) {
     this.twillSpecification = twillSpecification;
     this.fsUser = fsUser;
     this.twillAppDir = twillAppDir;
@@ -59,6 +60,7 @@ public class TwillRuntimeSpecification {
     this.logLevels = logLevels;
     this.maxRetries = maxRetries;
     this.minHeapRatio = minHeapRatio;
+    this.logCollectionEnabled = logCollectionEnabled;
   }
 
   public TwillSpecification getTwillSpecification() {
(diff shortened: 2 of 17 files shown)


twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
```diff
@@ -71,21 +71,37 @@
...
+        LOG.warn("Log collection is disabled for application {} with runId {}. " +
+                   "Adding log handler won't get any logs.", appName, runId);
+      }
+    }
   }
 
   @Override
   protected synchronized void doStartUp() {
-    if (!logHandlers.isEmpty()) {
+    if (kafkaClient != null && !logHandlers.isEmpty()) {
```
hsaputra: Should this just check `if (logCollectionEnabled)` instead since technically both conditions will be met when logCollectionEnabled is true

chtyim: `logCollectionEnabled` is not a field. I used `kafkaClient == null` to indicate disabling of log collection.
Also, the `logHandlers` collection can be empty during startup of the controller, even log collection is enabled, hence the check.

hsaputra: But `this.logHandlers` will always be empty if `if (kafkaClient == null)` is true? So could we just rely on check on kafkaClient as null ?

chtyim: you mean have the logic as

```java
if (kafkaClient == null) {
  return;
}
if (!logHandlers.isEmpty()) {
  kafkaClient.startAndWait();
  logCancellable = kafkaClient.getConsumer().....
}
```

??
Doesn't look simpler than the existing code to me.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillController.java
```diff
@@ -77,7 +77,7 @@
    */
   YarnTwillController(String appName, RunId runId, ZKClient zkClient,
                       final ApplicationMasterLiveNodeData amLiveNodeData, final YarnAppClient yarnAppClient) {
-    super(runId, zkClient, Collections.<LogHandler>emptyList());
+    super(appName, runId, zkClient, amLiveNodeData.getKafkaZKConnect() != null, Collections.<LogHandler>emptyList());
```
hsaputra: Can the constructor of `YarnTwillController` take new input of `logCollectionEnabled` ? I am seeing different ways to check whether to disable log or not

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -105,4 +114,17 @@ public String getRmSchedulerAddr() {
...
     return maxRetries;
   }
+
+  /**
+   * Returns the ZK connection string for the Kafka used for log collections,
+   * or {@code null} if log collection is disabled.
+   */
+  @Nullable
+  public String getKafkaZKConnect() {
+    if (!isLogCollectionEnabled()) {
```
hsaputra: I am not too keen about having this "side effect" check for log collection enabled here to return KafkaZKConnect.
Caller can always check `isLogCollectionEnabled()` to figure out whether log collection is disabled, right?

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillController.java
chtyim: ? I don't understand, can you give me an example? The current logic is, from an existing app discovered via ZK, if the AM live node data doesn't have the KafkaZKConnect string, that means the app was launched with log collection disabled, hence the logic here.

hsaputra:
I did a pass and add some comments. Mostly kind of concern of multiple ways to check whether log collection is disabled or not.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
chtyim: What do mean by side effect check? The contract of this method is to return `null` if log collection is disabled.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillController.java
chtyim: Or are you suggesting adding a new method `boolean isLogCollectionEnabled` to the `ApplicationMasterLiveNode` class?

chtyim:
@hsaputra Thanks for the review. Please see my replies.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
hsaputra: Thats actually what I meant. I am wondering if we could avoid returning null here because the extra check of if for `isLogCollectionEnabled`. The idea is the code that need to get KafkaZkconnect can get it when it is needed and dont call it when there is no need to do Kafka connect bc the caller can call `isLogCollectionEnabled` to determine whether to get the connect info or not.

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillController.java
hsaputra: Yeah, thinking about having `ApplicationMasterLiveNode.isLogCollectionEnabled` to make the contract more clear. Otherwise later we will forget why checking for getKafkaZKConnect != null here

twill-core/src/main/java/org/apache/twill/internal/AbstractTwillController.java
hsaputra: You are right here, there  is a chance that even log is enabled but caller can pass empty `logHandlers`.
+1 for current check

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillController.java
chtyim: The contract of the getKafkaZKConnect won't change though, since it has to return something (or throw exception). From the caller point of view, there is almost no differences.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
chtyim: How to avoid returning null? You mean by throwing exception so that the caller has to call isLogCollectionEnabled before calling this method? I am not a friend of method coupling like this.

hsaputra: The call to `getZkConnectStr` and `getTwillAppName` and `getTwillAppRunId` will never return null based on the contract (no @Nullable) , hence call to `getKafkaZKConnect` will never be null. 

I was saying that either way, someone need to check whether call to getKafkaZKConnect will return null if log collection is disabled, so why not ask to isLogCollectionEnabled to check first.

chtyim: This is because there is no way to enforce someone have to call `isLogCollectionEnabled` first before calling the `getKafkaZKConnect` method, meaning the `getKafkaZKConnect` method needs to handle the case when log collection disabled anyway (either by returning `null` or returning `Optional` or throwing exception). From the caller perspective, it seems more straight forward to call the method and validate against the contract then calling two methods in the right order.

However, having `getKafkaZKConnect` returning `null` (I prefer `Optional`, but since we are on Java 7, and we are on the road on reducing reliance on guava, so that's not really an option), can make cases like this simpler in future when adding support for TWILL-147. E.g.

```java
YarnTwillController(@Nullable String kafkaZKConnect, ...) {
  this.kafkaClient = createKafkaClient(kafkaZKConnect);
}

@Nullable
private KafkaClient createKafkaClient(@Nnullable String kafkaZKConnect) {
  if (kafkaZKConnect == null) {
    return null;
  }
  return new KafkaClient(....);
}

...
// For app launch with the current twill runner
new YarnTwillController(spec.getKafkaZKConnect(), ...);

// For running app recovered from ZK.
new YarnTwillController(amLiveNode.getKafkaZKConnect(), ...);
```

Without the `nullable` support, there would be more overloaded constructors as well as code branches to achieve the same.

hsaputra: Same thing of no way that caller could have guessed without looking at the code that there is a check of isLogCollectionEnabled inside getKafkaZKConnect that have inferred meaning when it is returning null.

The compromise I think at least added JavaDoc header for getKafkaZKConnect to explain why it could return null.

hsaputra:
+1 for correctness

I just added suggestion of style of return null vs declarative check by caller to determine if logging collection is disabled.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
chtyim: Do you mean this javadoc and the `@Nullable` annotation not enough?

```java
/**
  * Returns the ZK connection string for the Kafka used for log collections,
  * or {@code null} if log collection is disabled.
  */
@Nullable
public String getKafkaZKConnect()
```

hsaputra: Ah sorry, that will work, I was looking at snippet of change. Yeah, should work

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillController.java
hsaputra: At this point is more about style than correctness. I personally prefer to not make the logic of determining to enable or disable logging in YarnTwillController but rather on object that have current state of the union.

chtyim:
Thank @hsaputra for the review. Merging the change.

PR #42, opened 2017-03-27
(TWILL-171) Clone the HDFS delegation in HA mode.

chtyim:
- This is for working around HDFS-9276


chtyim:
@serranom True. We can also try to get it via the `LocationFactory` from the application location, however it won't really make the code easier to understand (there would be unwrapping and `instanceof` check)

anew:
IMO this is the cleaner, if verbose, way. LGTM +1

PR #43, opened 2017-03-27
(TWILL-194) Acquires KMS delegation token correctly

chtyim:
- This is to workaround HDFS-10296 that if FileContext object is used
  to acquire delegation token, the KMS delegation token is not included


anew:
LGTM

PR #47, opened 2017-03-28
(TWILL-223) Make FileContextLocationFactory UGI aware

chtyim:
- Use different FileContext object based on the caller UGI
- Allows sharing the same factory instance for different user

### DIFF
--- a/twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocationFactory.java
+++ b/twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocationFactory.java
@@ -18,13 +18,21 @@
 package org.apache.twill.filesystem;
 
 import com.google.common.base.Preconditions;
-import com.google.common.base.Throwables;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.UncheckedExecutionException;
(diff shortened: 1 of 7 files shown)


twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocationFactory.java
```diff
@@ -103,16 +138,32 @@ public Location create(URI uri) {
...
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to get current user information", e);
+    } catch (UncheckedExecutionException e) {
+      Throwable cause = e.getCause();
+      if (cause instanceof UnsupportedFileSystemException) {
+        String defaultURI = configuration.get(CommonConfigurationKeysPublic.FS_DEFAULT_NAME_KEY,
+                                              CommonConfigurationKeysPublic.FS_DEFAULT_NAME_DEFAULT);
+        throw new IllegalStateException("File system with URI '" + defaultURI + "' is not supported", cause);
+      }
+      throw new RuntimeException(cause);
```
albertshau: at this point the cause should be a RuntimeException right? in which case we should just propagate it directly rather than wrapping it.

chtyim: Well, it can be `Exception` or `RuntimeException`. I can add a check to just throw the cause without wrapping if the type is `RuntimeException`

PR #49, opened 2017-03-30
(TWILL-229) Default to use logback_template.xml to configure logger, …

chtyim:
…unless overridden by user


yaojiefeng:
Changes LGTM, may I ask where do we set the logback if it is overriden by user?

chtyim:
For AM, currently it is not settable. For Runnable containers, user can add as resources / localize a `logback.xml` file and specify it in the classpath.

yaojiefeng:
I see. LGTM

PR #51, opened 2017-04-03
(TWILL-179) Added support for custom ClassLoader for containers

chtyim:
- Added method TwillPreparer.setClassLoader
- Use system property "twill.custom.class.loader" to pass the class name of the custom ClassLoader

### DIFF
--- a/twill-yarn/src/test/java/org/apache/twill/yarn/CustomClassLoader.java
+++ b/twill-yarn/src/test/java/org/apache/twill/yarn/CustomClassLoader.java
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


twill-yarn/src/test/java/org/apache/twill/yarn/CustomClassLoader.java
```diff
@@ -0,0 +1,87 @@
...
+  protected Class<?> findClass(String name) throws ClassNotFoundException {
+    if (!CustomClassLoaderRunnable.GENERATED_CLASS_NAME.equals(name)) {
+      return super.findClass(name);
+    }
+
+    // Generate a class that look like this:
+    //
+    // public class Generated {
+    //
+    //   public void announce(ServiceAnnouncer announcer, String serviceName, int port) {
```
anwar6953: Is the javadoc outdated/inconsistent?
`Cancellable` is declared as the return type, below on line 76.

chtyim: Will update.

anwar6953:
1 minor comment; remainder LGTM.

twill-yarn/src/test/java/org/apache/twill/yarn/CustomClassLoader.java
chtyim: actually the comment is correct. The generated method has `void` return type. The `announcer.announce` method has `Cancellable` return type, but the result gets ignored (hence the `mg.pop()` code in line 79).

PR #53, opened 2017-04-03
(TWILL-230) Get resource report based on the caller user

chtyim:
- Also by default get the resource report from the tracking url, then fall back to the original tracking url.


twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillController.java
```diff
@@ -322,7 +314,46 @@ private boolean hasRun(YarnApplicationState state) {
...
   }
+
+  /**
+   * Returns the {@link ResourceReportClient} for fetching resource report from the AM.
+   * It first consults the RM for the tracking URL and get the resource report from there.
+   */
+  @Nullable
+  private ResourceReportClient getResourcesClient() {
+    YarnApplicationReport report = processController.getReport();
+    List<URL> urls = new ArrayList<>(2);
```
yaojiefeng: Why will this url list has initial capacity of 2?

chtyim: Because we store the tracking URL and the original tracking URL.

anwar6953:
LGTM

anwar6953:
LGTM

PR #55, opened 2017-06-24
(TWILL-237) Twill is using hdfs HAUtil api that is nont-compatible wi…

yufeldman:
…th hadoop 2.8

+ Use Java's MethodHandle (dynamic lang support) rather than Method (reflection) in FileContextLocationUtil
+ Expose static API rather than the handle itself

### DIFF
--- a/twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocation.java
+++ b/twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocation.java
@@ -19,6 +19,7 @@ package org.apache.twill.filesystem;
 
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
+
 import org.apache.hadoop.fs.CreateFlag;
 import org.apache.hadoop.fs.FileAlreadyExistsException;
 import org.apache.hadoop.fs.FileContext;
@@ -28,7 +29,6 @@ import org.apache.hadoop.fs.ParentNotDirectoryException;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.RemoteIterator;
 import org.apache.hadoop.fs.permission.FsPermission;
-import org.apache.hadoop.hdfs.HAUtil;
 import org.apache.hadoop.security.AccessControlException;
 
 import java.io.FileNotFoundException;
@@ -41,6 +41,7 @@ import java.util.EnumSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
+
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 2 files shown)


twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocation.java
```diff
@@ -162,7 +163,8 @@ public URI toURI() {
     // append "port" to the path URI, while the DistributedFileSystem always use the cluster logical
     // name, which doesn't allow having port in it.
     URI uri = path.toUri();
-    if (HAUtil.isLogicalUri(locationFactory.getConfiguration(), uri)) {
+
+    if (FileContextLocationUtil.useLogicalUri(locationFactory.getConfiguration(), uri)) {
```
chtyim: It seems like we actually don't need to check, but rather just always strip off the port from the `URI`, since the `FileContext` or `FileSystem` in Hadoop will always determine it internally based on the `Configuration`.

In fact that would make the returning `URI` more predictable.

yufeldman: Should we just get rid of the check in this case?

chtyim: Yes. I think it should be safe to remove and always return an URI without port

yufeldman: I am not sure it is as simple as that. We may need to change equals() to compare based on toURI() and not path or somehow make path and toURI() to be in sync within that class, otherwise parts become inconsistent and I am not sure unitTests failures in this case is the only issue.

chtyim: That make sense.

twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocationUtil.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.net.URI;
+
+/**
+ * Utility class.
+ */
+class FileContextLocationUtil {
```
chtyim: Util class should be `final`.

twill-yarn/src/main/java/org/apache/twill/filesystem/FileContextLocationUtil.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.net.URI;
+
+/**
+ * Utility class.
+ */
+final class FileContextLocationUtil {
```
yufeldman: Thank you for the final review. Addressed your comment

PR #56, opened 2017-07-08
(TWILL-238) Change TwillController restartInstances API

fonelor:
Implemented https://issues.apache.org/jira/browse/TWILL-238 

New method in TwillController to restart instances of runnable

### DIFF
--- a/twill-api/src/main/java/org/apache/twill/api/TwillController.java
+++ b/twill-api/src/main/java/org/apache/twill/api/TwillController.java
@@ -17,15 +17,14 @@
  */
 package org.apache.twill.api;
 
-import org.apache.twill.api.logging.LogEntry;
-import org.apache.twill.api.logging.LogHandler;
-import org.apache.twill.discovery.Discoverable;
-import org.apache.twill.discovery.ServiceDiscovered;
-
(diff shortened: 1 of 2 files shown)


fonelor:
Thank you, serranom, for your comment. Is there any way to set "right" order of imports in IDEA? I've imported settings from HowToContribute page, but it seems, that it still wants to move imports.

twill-api/src/main/java/org/apache/twill/api/TwillController.java
```diff
@@ -17,15 +17,14 @@
...
-import org.apache.twill.api.logging.LogEntry;
-import org.apache.twill.api.logging.LogHandler;
-import org.apache.twill.discovery.Discoverable;
-import org.apache.twill.discovery.ServiceDiscovered;
-
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Future;
 import javax.annotation.Nullable;
+import org.apache.twill.api.logging.LogEntry;
```
fonelor: i've reverted reformatting changes

fonelor:
Reverted reformatting

chtyim:
LGTM.

PR #58, opened 2017-07-26
[TWILL-240] EventHandler Improvement

maochf:
https://issues.apache.org/jira/browse/TWILL-240

Add started, containerLaunched, containerStopped, completed, killed, aborted to EventHandler

### DIFF
--- a/twill-api/src/main/java/org/apache/twill/api/EventHandler.java
+++ b/twill-api/src/main/java/org/apache/twill/api/EventHandler.java
@@ -17,6 +17,8 @@
  */
 package org.apache.twill.api;
 
+import org.apache.twill.internal.Constants;
+
 import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
@@ -124,6 +126,58 @@ public abstract class EventHandler {
   }
 
   /**
+   * Invoked by the application when it starts.
+   */
+  public void started() {
+    // No-op
+  }
+
+  /**
+   * Invoked by the application when new container is launched for a {@link TwillRunnable}.
+   *
+   * @param runnableName name of the runnable to be run in the new container
+   * @param instanceId the instance ID of the runnable instance to be run in the new container
+   * @param containerId the ID of the newly launched container
+   */
+  public void containerLaunched(String runnableName, int instanceId, String containerId) {
+    // No-op
+  }
+
+  /**
+   * Invoked by the application when the container allocated for a {@link TwillRunnable} is stopped.
+   *
+   * @param runnableName name of the runnable in the stopped container
+   * @param instanceId the instance ID of the runnable instance run in the stopped container
+   * @param containerId the ID of the stopped container
+   * @param exitStatus the exit status of the container
+   */
+  public void containerStopped(String runnableName, int instanceId, String containerId, int exitStatus) {
+    // No-op
+  }
+
+  /**
+   * Invoked by the application when all containers complete.
+   */
+  public void completed() {
+    // No-op
+  }
+
+  /**
+   * Invoked by the application when stop command is received to kill the current application.
+   *
+   */
+  public void killed() {
+    // No-op
+  }
+
+  /**
+   * Invoked by the application when the application is aborted because of timeout.
+   */
+  public void aborted() {
+    // No-op
+  }
+
+  /**
    * Invoked by the application when shutting down.
    */
   public void destroy() {
@@ -133,9 +187,11 @@ public abstract class EventHandler {
   /**
    * Invoked when the number of expected instances doesn't match with number of actual instances.
    * @param timeoutEvents An Iterable of {@link TimeoutEvent} that contains information about runnable launch timeout.
-   * @return A {@link TimeoutAction} to govern action to act.
+   * @return A {@link TimeoutAction} to govern action to act. By default it is to recheck after 60 seconds.
    */
-  public abstract TimeoutAction launchTimeout(Iterable<TimeoutEvent> timeoutEvents);
+  public TimeoutAction launchTimeout(Iterable<TimeoutEvent> timeoutEvents) {
+    return new TimeoutAction(Constants.PROVISION_TIMEOUT);
+  }
 
   /**
    * Returns set of configurations available at runtime for access.
--- a/twill-api/src/main/java/org/apache/twill/api/EventHandlerContext.java
+++ b/twill-api/src/main/java/org/apache/twill/api/EventHandlerContext.java
@@ -22,5 +22,18 @@ package org.apache.twill.api;
  */
 public interface EventHandlerContext {
 
+  /**
+   * @return Name of the {@link TwillApplication} containing the {@link EventHandler}
+   */
+  String getApplicationName();
+
+  /**
+   * @return the unique ID of the current run of the {@link TwillApplication} containing the {@link EventHandler}
+   */
+  RunId getRunId();
+
+  /**
+   * @return {@link EventHandlerSpecification} of the {@link EventHandler}
+   */
   EventHandlerSpecification getSpecification();
 }
--- a/twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
+++ b/twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
@@ -46,6 +46,7 @@ import org.apache.hadoop.yarn.api.records.Resource;
 import org.apache.hadoop.yarn.util.Records;
 import org.apache.twill.api.Command;
 import org.apache.twill.api.EventHandler;
+import org.apache.twill.api.EventHandlerContext;
 import org.apache.twill.api.EventHandlerSpecification;
 import org.apache.twill.api.LocalFile;
 import org.apache.twill.api.ResourceReport;
@@ -107,6 +108,13 @@ import javax.annotation.Nullable;
  * The class that acts as {@code ApplicationMaster} for Twill applications.
  */
 public final class ApplicationMasterService extends AbstractYarnTwillService implements Supplier<ResourceReport> {
+  /**
+   * Final status of this service when it stops.
+   */
+  private enum StopStatus {
+    COMPLETED, // All containers complete
+    ABORTED // Aborted because of timeout
+  }
 
   private static final Logger LOG = LoggerFactory.getLogger(ApplicationMasterService.class);
   private static final Gson GSON = new GsonBuilder()
@@ -131,6 +139,7 @@ public final class ApplicationMasterService extends AbstractYarnTwillService imp
   private final Map<String, Map<String, String>> environments;
   private final TwillRuntimeSpecification twillRuntimeSpec;
 
+  private volatile StopStatus stopStatus;
   private volatile boolean stopped;
   private Queue<RunnableContainerRequest> runnableContainerRequests;
   private ExecutorService instanceChangeExecutor;
@@ -157,8 +166,8 @@ public final class ApplicationMasterService extends AbstractYarnTwillService imp
                                                         twillRuntimeSpec.getKafkaZKConnect());
 
     this.expectedContainers = new ExpectedContainers(twillSpec);
-    this.runningContainers = createRunningContainers(amClient.getContainerId(), amClient.getHost());
     this.eventHandler = createEventHandler(twillSpec);
+    this.runningContainers = createRunningContainers(amClient.getContainerId(), amClient.getHost());
   }
 
   private JvmOptions loadJvmOptions() throws IOException {
@@ -172,19 +181,104 @@ public final class ApplicationMasterService extends AbstractYarnTwillService imp
   }
 
   @SuppressWarnings("unchecked")
-  @Nullable
   private EventHandler createEventHandler(TwillSpecification twillSpec) throws ClassNotFoundException {
     // Should be able to load by this class ClassLoader, as they packaged in the same jar.
     EventHandlerSpecification handlerSpec = twillSpec.getEventHandler();
     if (handlerSpec == null) {
-      return null;
+      // if no handler is specified, return an EventHandler with no-op
+      return new EventHandler() {};
     }
 
     Class<?> handlerClass = getClass().getClassLoader().loadClass(handlerSpec.getClassName());
     Preconditions.checkArgument(EventHandler.class.isAssignableFrom(handlerClass),
                                 "Class {} does not implements {}",
                                 handlerClass, EventHandler.class.getName());
-    return Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    final EventHandler delegate = Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    // wrap all calls to the delegate EventHandler methods except initialize so that all errors will be caught
+    return new EventHandler() {
+
+      @Override
+      public void initialize(EventHandlerContext context) {
+        delegate.initialize(context);
+      }
+
+      @Override
+      public void started() {
+        try {
+          delegate.started();
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.started()", delegate.getClass().getName(), t);
+        }
+      }
+
+      @Override
+      public void containerLaunched(String runnableName, int instanceId, String containerId) {
+        try {
+          delegate.containerLaunched(runnableName, instanceId, containerId);
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.containerLaunched(String, int, String)",
+                   delegate.getClass().getName(), t);
+        }
+      }
+
+      @Override
+      public void containerStopped(String runnableName, int instanceId, String containerId, int exitStatus) {
+        try {
+          delegate.containerStopped(runnableName, instanceId, containerId, exitStatus);
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.containerStopped(String, int, String, int)",
+                   delegate.getClass().getName(), t);
+        }
+      }
+
+      @Override
+      public void completed() {
+        try {
+          delegate.completed();
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.completed()", delegate.getClass().getName(), t);
+        }
+      }
+
+      @Override
+      public void killed() {
+        try {
+          delegate.killed();
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.killed()", delegate.getClass().getName(), t);
+        }
+      }
+
+      @Override
+      public void aborted() {
+        try {
+          delegate.aborted();
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.aborted()", delegate.getClass().getName(), t);
+        }
+      }
+
+      @Override
+      public void destroy() {
+        try {
+          delegate.destroy();
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.destroy()", delegate.getClass().getName(), t);
+        }
+      }
+
+      @Override
+      public TimeoutAction launchTimeout(Iterable<TimeoutEvent> timeoutEvents) {
+        try {
+          return delegate.launchTimeout(timeoutEvents);
+        } catch (Throwable t) {
+          LOG.warn("Exception raised when calling {}.launchTimeout(Iterable<TimeoutEvent>)",
+                   delegate.getClass().getName(), t);
+        }
+        // call super.launchTimeout in case of any errors from the delegate
+        return super.launchTimeout(timeoutEvents);
+      }
+    };
   }
 
   private RunningContainers createRunningContainers(ContainerId appMasterContainerId,
@@ -203,8 +297,8 @@ public final class ApplicationMasterService extends AbstractYarnTwillService imp
       maxHeapMemoryMB,
       appMasterHost, null);
     String appId = appMasterContainerId.getApplicationAttemptId().getApplicationId().toString();
-    return new RunningContainers(appId, appMasterResources, zkClient, applicationLocation,
-      twillSpec.getRunnables(), twillRuntimeSpec.getMaxRetries());
+    return new RunningContainers(twillRuntimeSpec, appId, appMasterResources, zkClient, applicationLocation,
+                                 twillSpec.getRunnables(), eventHandler);
   }
 
   @Override
@@ -218,9 +312,9 @@ public final class ApplicationMasterService extends AbstractYarnTwillService imp
       TwillRuntimeSpecificationAdapter.create().toJson(twillRuntimeSpec));
 
     // initialize the event handler, if it fails, it will fail the application.
-    if (eventHandler != null) {
-      eventHandler.initialize(new BasicEventHandlerContext(twillSpec.getEventHandler()));
-    }
+    eventHandler.initialize(new BasicEventHandlerContext(twillRuntimeSpec));
+    // call event handler started.
+    eventHandler.started();
 
     instanceChangeExecutor = Executors.newSingleThreadExecutor(Threads.createDaemonThreadFactory("instanceChanger"));
 
@@ -236,15 +330,6 @@ public final class ApplicationMasterService extends AbstractYarnTwillService imp
 
     LOG.info("Stop application master with spec: {}",
              TwillRuntimeSpecificationAdapter.create().toJson(twillRuntimeSpec));
-
-    if (eventHandler != null) {
-      try {
-        // call event handler destroy. If there is error, only log and not affected stop sequence.
-        eventHandler.destroy();
-      } catch (Throwable t) {
-        LOG.warn("Exception when calling {}.destroy()", eventHandler.getClass().getName(), t);
-      }
-    }
 
     instanceChangeExecutor.shutdownNow();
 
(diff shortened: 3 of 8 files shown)


twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -267,6 +281,9 @@ void stopByIdAndWait(String runnableName, int instanceId) {
 
         resourceReport.removeRunnableResources(runnableName, containerId);
         containerChange.signalAll();
+        if (eventHandler != null) {
+          eventHandler.containerLaunched(twillAppName, runId, runnableName, instanceId, containerId);
```
sameetandpotatoes: Typo - should be `containerStopped`

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -103,6 +105,8 @@ public Integer apply(BitSet input) {
   private final Table<String, String, TwillContainerController> containers;
 
   // Map from runnableName to a BitSet, with the <instanceId> bit turned on for having an instance running.
+  private final String twillAppName;
```
sameetandpotatoes: The comment above this line should be moved down to above the `runnableInstances` variable

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -222,6 +226,7 @@ protected void doStart() throws Exception {
     // initialize the event handler, if it fails, it will fail the application.
     if (eventHandler != null) {
       eventHandler.initialize(new BasicEventHandlerContext(twillSpec.getEventHandler()));
+      eventHandler.started(twillRuntimeSpec.getTwillAppName(), twillRuntimeSpec.getTwillAppRunId());
```
sameetandpotatoes: According to the comment above this `if` statement, if the event handler fails to initialize, will any twill application state be recorded?

maochf: If initialization fails, then the whole app will fail. `TwillController` should be able to catch this

twill-api/src/main/java/org/apache/twill/api/EventHandler.java
```diff
@@ -124,6 +124,74 @@ public void initialize(EventHandlerContext context) {
...
+  /**
+   * Invoked by the application when a container is stopped.
+   *
+   * @param twillAppName name of the current application
+   * @param runId run ID of current application run
+   * @param runnableName name of the runnable in the stopped container
+   * @param instanceId the instance ID of the runnable instance run in the stopped container
+   * @param containerId the ID of the stopped container
+   */
+  public void containerStopped(String twillAppName, RunId runId, String runnableName,
```
chtyim: Should provide the exit code as well.

twill-api/src/main/java/org/apache/twill/api/EventHandler.java
```diff
@@ -124,6 +124,74 @@ public void initialize(EventHandlerContext context) {
...
+   * @param runId run ID of current application run
+   */
+  public void aborted(String twillAppName, RunId runId) {
+    // No-op
+  }
+
+  /**
    * Invoked by the application when shutting down.
    */
   public void destroy() {
```
chtyim: Can't add comment at line 206, but it's better to make the `launchTimeout` method non-abstract.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -135,6 +135,9 @@
   private final Map<String, Map<String, String>> environments;
   private final TwillRuntimeSpecification twillRuntimeSpec;
 
+  private volatile boolean completed;
```
chtyim: Instead of having three variables for `completed`, `aborted`, `stopped`, it's better to have one Enum to represent them.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -205,8 +208,9 @@ private RunningContainers createRunningContainers(ContainerId appMasterContainer
       Integer.parseInt(System.getenv(EnvKeys.YARN_CONTAINER_MEMORY_MB)),
       appMasterHost, null);
     String appId = appMasterContainerId.getApplicationAttemptId().getApplicationId().toString();
-    return new RunningContainers(appId, appMasterResources, zkClient, applicationLocation,
-      twillSpec.getRunnables(), twillRuntimeSpec.getMaxRetries());
+    return new RunningContainers(twillRuntimeSpec.getTwillAppName(), twillRuntimeSpec.getTwillAppRunId(),
```
chtyim: Better just pass the `twillRuntimeSpec` to the `RunningContainers` instead of getting individual entries one by one. It can replace the `twillRuntimeSpec.getXXX` and `twillSpec.getRunnables`.

twill-api/src/main/java/org/apache/twill/api/EventHandler.java
maochf: what should be the default action of `launchTimeout`?

chtyim: continue, with the recheck time to some hardcode constant (say 60 seconds).

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -135,6 +135,15 @@
   private final Map<String, Map<String, String>> environments;
   private final TwillRuntimeSpecification twillRuntimeSpec;
 
+  /**
+   * Final status of this service when it stops.
+   */
+  public enum FinalStatus {
```
chtyim: Sounds more like `StopStatus` or `CompletionStatus`. Also it should be `private` since it is only used in this class.

Also, please move it before all fields declaration instead of having it in between fields.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -291,6 +301,23 @@ public void run() {
...
+      switch (finalStatus) {
+        case COMPLETED:
+          eventHandler.completed(twillRuntimeSpec.getTwillAppName(), twillRuntimeSpec.getTwillAppRunId());
+          break;
+        case ABORTED:
+        eventHandler.aborted(twillRuntimeSpec.getTwillAppName(), twillRuntimeSpec.getTwillAppRunId());
+          break;
+        default:
+          // should never reach here
+          LOG.error("Unsupported FinalStatus '%s'", finalStatus.name());
```
chtyim: That's not the right syntax. slf4j logger uses `{}`. Should be `LOG.error("Unsupported status {}", finalStatus)`.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -291,6 +301,23 @@ public void run() {
     // Since all the runnables are now stopped, it is okay to stop the poller.
     stopPoller.shutdownNow();
     cleanupDir();
+    if (eventHandler != null) {
+      if (finalStatus == null) {
+        // if finalStatus is not set, the application must be stopped by a SystemMessages#STOP_COMMAND
+        eventHandler.killed(twillRuntimeSpec.getTwillAppName(), twillRuntimeSpec.getTwillAppRunId());
+      }
+      switch (finalStatus) {
```
chtyim: This may have NPE. Should have a `else`.

twill-api/src/main/java/org/apache/twill/api/EventHandler.java
```diff
@@ -124,6 +124,75 @@ public void initialize(EventHandlerContext context) {
   }
 
   /**
+   * Invoked by the application when it starts.
+   *
+   * @param twillAppName name of the current application
+   * @param runId run ID of current application run
+   */
+  public void started(String twillAppName, RunId runId) {
```
chtyim: Why all methods has the app name and runid? It should be provided via the `EventHandlerContext` since it will never change for an application run.

twill-yarn/src/test/java/org/apache/twill/yarn/ProvisionTimeoutTestRun.java
```diff
@@ -79,12 +83,22 @@ public void initialize(EventHandlerContext context) {
 
     @Override
     public TimeoutAction launchTimeout(Iterable<TimeoutEvent> timeoutEvents) {
+      for (TimeoutEvent event : timeoutEvents) {
+        LOG.info("Requested {} containers for runnable {}, only got {} after {} ms.",
```
chtyim: What's the point of this logging?

twill-yarn/src/test/java/org/apache/twill/yarn/ProvisionTimeoutTestRun.java
```diff
@@ -79,12 +83,22 @@ public void initialize(EventHandlerContext context) {
...
       if (abort) {
         return TimeoutAction.abort();
       } else {
         return TimeoutAction.recheck(10, TimeUnit.SECONDS);
       }
     }
+
+    @Override
+    public void aborted(String twillAppName, RunId runId) {
+      LOG.info(String.format("Aborted %s with runId %s", twillAppName, runId.getId()));
```
chtyim: Besides logging, can you add/modify the test to actually validate the new lifecycle methods in the EventHandler?

twill-yarn/src/test/java/org/apache/twill/yarn/ProvisionTimeoutTestRun.java
maochf: i was using it to confirm the change. Will remove

twill-api/src/main/java/org/apache/twill/api/EventHandler.java
maochf: the `EventHandlerContext` just contains a `EventHandlerSpecification`, and it doesn't seem to guarantee app name and runid are there?

twill-api/src/main/java/org/apache/twill/api/EventHandlerContext.java
```diff
@@ -22,5 +22,9 @@
  */
 public interface EventHandlerContext {
 
+  String getTwillAppName();
```
chtyim: Should be called `getApplicationName()`.

twill-api/src/main/java/org/apache/twill/api/EventHandlerContext.java
```diff
@@ -22,5 +22,9 @@
...
+
+  String getTwillAppRunId();
```
chtyim: Just call it `getRunId()`.

twill-api/src/main/java/org/apache/twill/api/EventHandlerContext.java
chtyim: Also please add javadoc.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -291,6 +300,24 @@ public void run() {
...
+        // if finalStatus is not set, the application must be stopped by a SystemMessages#STOP_COMMAND
+        eventHandler.killed(twillRuntimeSpec.getTwillAppName(), twillRuntimeSpec.getTwillAppRunId());
+        return;
+      }
+      switch (stopStatus) {
+        case COMPLETED:
+          eventHandler.completed();
+          break;
+        case ABORTED:
+        eventHandler.aborted();
```
chtyim: Misalignment.

twill-api/src/main/java/org/apache/twill/api/EventHandlerContext.java
chtyim: The run id should be of type `RunId`.

twill-yarn/src/test/java/org/apache/twill/yarn/ProvisionTimeoutTestRun.java
```diff
@@ -85,13 +105,29 @@ public TimeoutAction launchTimeout(Iterable<TimeoutEvent> timeoutEvents) {
         return TimeoutAction.recheck(10, TimeUnit.SECONDS);
       }
     }
+
+    @Override
+    public void aborted() {
+      try {
+        new File(context.getSpecification().getConfigs().get("parentFolderPath") + File.separator
```
chtyim: When constructing file, usually not to use `File.separator`, but instead use the constructor `File(File parent, String path)`.

chtyim: Also, we are only testing abort? How about other lifecycle methods?

twill-yarn/src/test/java/org/apache/twill/yarn/EventHandlerTest.java
```diff
@@ -0,0 +1,351 @@
...
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Tests {@link EventHandler} methods
+ */
+public final class EventHandlerTest extends BaseYarnTest {
```
chtyim: Rename the test class to `EventHandlerTestRun` and add it to `YarnTestSuite`.

twill-yarn/src/test/java/org/apache/twill/yarn/EventHandlerTest.java
```diff
@@ -0,0 +1,351 @@
...
+      // for the same number of times
+      String[] containerLaunchedFiles = new File(parentFolder.getAbsolutePath(), CONTAINER_LAUNCHED_FOLDER).list();
+      String[] containerStoppedFiles = new File(parentFolder.getAbsolutePath(), CONTAINER_STOPPED_FOLDER).list();
+      Assert.assertEquals(containerLaunchedFiles.length, containerStoppedFiles.length);
+      Assert.assertTrue(Arrays.asList(containerLaunchedFiles).containsAll(Arrays.asList(containerStoppedFiles)));
+      // EventHandler#completed() method should be called to create a file
+      Assert.assertTrue(handlerFiles.contains(COMPLETED_FILE));
+    } catch (Exception e) {
+      // kill the app as cleanup
+      controller.kill();
```
chtyim: This is not needed. The base test class will handle it.

twill-yarn/src/test/java/org/apache/twill/yarn/EventHandlerTest.java
```diff
@@ -0,0 +1,351 @@
...
+      Set<String> handlerFiles = new HashSet<>(Arrays.asList(parentFolder.list()));
+      // EventHandler#killed() method should be called to create a file
+      Assert.assertTrue(handlerFiles.contains(KILLED_FILE));
+      // EventHandler#completed() method should not be called
+      Assert.assertFalse(handlerFiles.contains(COMPLETED_FILE));
+      // EventHandler#aborted() method should not be called
+      Assert.assertFalse(handlerFiles.contains(ABORTED_FILE));
+    } catch (Exception e) {
+      // kill the app as cleanup
+      controller.kill();
```
chtyim: Same as above.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -221,7 +229,8 @@ protected void doStart() throws Exception {
 
     // initialize the event handler, if it fails, it will fail the application.
     if (eventHandler != null) {
-      eventHandler.initialize(new BasicEventHandlerContext(twillSpec.getEventHandler()));
+      eventHandler.initialize(new BasicEventHandlerContext(twillRuntimeSpec, twillSpec.getEventHandler()));
```
chtyim: The `twillRuntimeSpec` already contains `twillSpec`. Seems unnecessary to pass in `twillSpec.getEventHandler()` as a separate parameter.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -173,6 +183,9 @@ void start(String runnableName, ContainerInfo containerInfo, TwillContainerLaunc
         startSequence.addLast(runnableName);
       }
       containerChange.signalAll();
+      if (eventHandler != null) {
+        eventHandler.containerLaunched(runnableName, instanceId, containerInfo.getId());
```
chtyim: How do we handle exception raised from event handler methods?

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -267,6 +280,14 @@ void stopByIdAndWait(String runnableName, int instanceId) {
 
         resourceReport.removeRunnableResources(runnableName, containerId);
         containerChange.signalAll();
+        if (eventHandler != null) {
+          Integer exitStatus = containerExitStatus.get(containerId);
```
chtyim: Should remove it from the exit status map, otherwise the map can keep growing in size.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -267,6 +280,14 @@ void stopByIdAndWait(String runnableName, int instanceId) {
...
+          if (exitStatus == null) {
```
chtyim: So this is for handling the case when `handleCompleted` was not called (e.g. the runnable container is not stopping after receiving the "stop" command), right? If that's the case, isn't that the exit status is always not there?

Also, do we need to do the similar logic to call event handler in the `stopAll` method?

twill-yarn/src/test/java/org/apache/twill/yarn/EventHandlerTest.java
```diff
@@ -0,0 +1,351 @@
...
+
+    private void createContainerFile(String runnableName, int instanceId, String containerId, String folderKey) {
+      Map<String, String> configs = context.getSpecification().getConfigs();
+      File launchedFolder = new File(configs.get("parentFolderPath"), configs.get(folderKey));
+      if (!launchedFolder.exists()) {
+        launchedFolder.mkdirs();
+        launchedFolder.setReadable(true, false);
+      }
+      try {
+        new File(launchedFolder.getAbsolutePath(), Joiner.on(":").join(runnableName, instanceId, containerId))
```
chtyim: Would be clear if you have a private `touch(String name)` method rather than have `try-catch` block on each callback method.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
maochf: `controller.stopAndWait();` at line 269 blocks until `handleCompleted` method runs

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -181,7 +189,6 @@ public Reader getInput() throws IOException {
   }
 
   @SuppressWarnings("unchecked")
-  @Nullable
   private EventHandler createEventHandler(TwillSpecification twillSpec) throws ClassNotFoundException {
     // Should be able to load by this class ClassLoader, as they packaged in the same jar.
     EventHandlerSpecification handlerSpec = twillSpec.getEventHandler();
```
chtyim: if `handlerSpec` is `null`, you still want to return a no-op handler, right?

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -193,7 +200,91 @@ private EventHandler createEventHandler(TwillSpecification twillSpec) throws Cla
     Preconditions.checkArgument(EventHandler.class.isAssignableFrom(handlerClass),
                                 "Class {} does not implements {}",
                                 handlerClass, EventHandler.class.getName());
-    return Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    final EventHandler delegate = Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    if (delegate == null) {
```
chtyim: `Instances.newInstance` never return `null`.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -193,7 +200,91 @@ private EventHandler createEventHandler(TwillSpecification twillSpec) throws Cla
...
+    }
+    // wrap the delegate EventHandler so that all errors will be caught
+    return new EventHandler() {
+
+      @Override
+      public void started() {
+        try {
+          delegate.started();
+        } catch (Throwable t) {
+          LOG.warn("Exception when calling {}.started()", eventHandler.getClass().getName(), t);
```
chtyim: Shouldn't it be `delegate.getClass().getName()`??

chtyim: `Exception raised when...`

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -193,7 +200,91 @@ private EventHandler createEventHandler(TwillSpecification twillSpec) throws Cla
...
                                 "Class {} does not implements {}",
                                 handlerClass, EventHandler.class.getName());
-    return Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    final EventHandler delegate = Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    if (delegate == null) {
+      // if no handler is specified, return an EventHandler with no-op
+      return new EventHandler() {};
+    }
+    // wrap the delegate EventHandler so that all errors will be caught
+    return new EventHandler() {
```
chtyim: You'll need to wrap the `initialize` as well. Also, I'd suggest if `initialize` throw exception, you'll just ignore all calls to other lifecycle methods to avoid excessive logging.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/RunningContainers.java
```diff
@@ -401,6 +409,24 @@ void stopAll() {
...
+          boolean containerExist = false;
+          for (ContainerInfo containerInfo : containerInfos) {
+            if (containerInfo.getId().equals(containerControllerEntry.getKey())) {
+              containerExist = true;
+              break;
+            }
+          }
+          // Only call eventHandler.containerStopped if container is not removed by handleCompleted
+          if (containerExist) {
+            eventHandler.containerStopped(runnableName, containerControllerEntry.getValue().getInstanceId(),
```
chtyim: you can move this inside the `for` loop and no need to use `containerExist`.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
maochf: originally, if `EventHandler#initialize` fails, the app will fail to start. Do we want to change this behavior?

chtyim: hum.. that's true.. maybe we should keep it.

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -181,19 +190,130 @@ public Reader getInput() throws IOException {
...
-    return Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    final EventHandler delegate = Instances.newInstance((Class<? extends EventHandler>) handlerClass);
+    // wrap all calls to the delegate EventHandler methods except initialize so that all errors will be caught
+    return new EventHandler() {
+      private boolean initialized;
+
+      @Override
+      public void initialize(EventHandlerContext context) {
+        delegate.initialize(context);
+        initialized = true;
```
chtyim: Since any exception thrown from `initialize` will terminate the app, I think we don't need this, right?

twill-yarn/src/main/java/org/apache/twill/internal/appmaster/ApplicationMasterService.java
```diff
@@ -239,14 +359,8 @@ protected void doStop() throws Exception {
...
-    if (eventHandler != null) {
-      try {
-        // call event handler destroy. If there is error, only log and not affected stop sequence.
-        eventHandler.destroy();
-      } catch (Throwable t) {
-        LOG.warn("Exception when calling {}.destroy()", eventHandler.getClass().getName(), t);
-      }
-    }
+    // call event handler destroy
+    eventHandler.destroy();
```
chtyim: Shouldn't call `destroy()` here anymore. It should be called after call other calls to the event handler, as it is acted as a cleanup call.

maochf: call it at the end of `doStop() `?

chtyim: I think so.

chtyim:
LGTM

maochf:
squashed

PR #59, opened 2017-08-05
(TWILL-241) Added per runnable configuration and jvm options support

chtyim:
This PR has two commits, one for adding per runnable configuration, the other for adding per runnable jvm options.


twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
```diff
@@ -183,6 +184,13 @@ public TwillPreparer withConfiguration(Map<String, String> config) {
   }
 
   @Override
+  public TwillPreparer withConfiguration(String runnableName, Map<String, String> config) {
```
anew: If the config contains only one key, say reserved memory, will the remaining keys default to the values configured for the entire app? Or will they default to the system default?

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -87,19 +86,46 @@ public String getTwillAppName() {
...
+   */
+  public double getAMMinHeapRatio() {
+    return getMinHeapRatio(config);
+  }
+
+  /**
+   * Returns the minimum heap ratio for the given runnable.
+   */
+  public double getMinHeapRatio(String runnableName) {
+    return getMinHeapRatio(runnableConfigs.containsKey(runnableName) ? runnableConfigs.get(runnableName) : config);
```
anew: I think this logic is incorrect. It might be that the runnable has its own config but does specify a min heap ratio. In that case the min heap ratio should come from the (appllication) config, and only if that does not specify a min heap ratio, use the default.

twill-yarn/src/test/java/org/apache/twill/yarn/ContainerSizeTestRun.java
```diff
@@ -67,20 +69,37 @@ public void testContainerSize() throws InterruptedException, TimeoutException, E
...
+
     TwillController controller = runner.prepare(new MaxHeapApp())
-      // Alter the AM container size
-      .withConfiguration(Collections.singletonMap(Configs.Keys.YARN_AM_MEMORY_MB, "256"))
+      // Alter the AM container size and heap ratio
+      .withConfiguration(ImmutableMap.of(Configs.Keys.YARN_AM_MEMORY_MB, "256",
+                                         Configs.Keys.HEAP_RESERVED_MIN_RATIO, "0.65"))
+      // Use a different heap ratio and reserved memory size for the runnable
+      .withConfiguration(runnableName,
+                         ImmutableMap.of(Configs.Keys.HEAP_RESERVED_MIN_RATIO, "0.8",
```
anew: can you add a test that, if this config for the runnable does not have HEAP_RESERVED_MIN_RATIO, it defaults to the one above (0.65) and not the system default.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
chtyim: Agree. Get it wrong after some refactoring

twill-yarn/src/main/java/org/apache/twill/yarn/YarnTwillPreparer.java
chtyim: Should default to the app setting first, then to the constants default

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -171,9 +168,27 @@ public String getKafkaZKConnect() {
...
-      Configs.Defaults.HEAP_RESERVED_MIN_RATIO;
+  private double getMinHeapRatio(@Nullable Map<String, String> config, double defaultValue) {
+    if (config == null) {
+      return defaultValue;
+    }
+
+    double ratio = config.containsKey(Configs.Keys.HEAP_RESERVED_MIN_RATIO) ?
+      Double.parseDouble(config.get(Configs.Keys.HEAP_RESERVED_MIN_RATIO)) : defaultValue;
+    // Ratio can't be <= 0
+    return ratio <= 0d ? defaultValue : ratio;
```
anew: It would be good to log a warning if it is <0, because that is a configuration error.

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -171,9 +168,27 @@ public String getKafkaZKConnect() {
   /**
    * Returns the minimum heap ratio ({@link Configs.Keys#HEAP_RESERVED_MIN_RATIO}) based on the given configuration.
    */
-  private double getMinHeapRatio(Map<String, String> config) {
-    return config.containsKey(Configs.Keys.HEAP_RESERVED_MIN_RATIO) ?
-      Double.parseDouble(config.get(Configs.Keys.HEAP_RESERVED_MIN_RATIO)) :
-      Configs.Defaults.HEAP_RESERVED_MIN_RATIO;
+  private double getMinHeapRatio(@Nullable Map<String, String> config, double defaultValue) {
+    if (config == null) {
```
anew: I think this would be easier to read, and it would give better errors, as follows:
```
if (config == null || !config.containsKey(...)) {
  return defaultValue;
} 
try {
  double ratio = Double.parseDouble(config.get(...));
} catch (NumberFormatException) {
  // either warn or throw a more informative exn (including runnable name and the bad value)
}
if (ratio <= 0d) {
  // either warn or throw a more informative exn (including runnable name and the bad value)
}
return ratio;
```

twill-core/src/main/java/org/apache/twill/internal/TwillRuntimeSpecification.java
```diff
@@ -171,9 +168,27 @@ public String getKafkaZKConnect() {
...
+    // Ratio can't be <= 0
+    return ratio <= 0d ? defaultValue : ratio;
+  }
+
+  /**
+   * Returns the reserved memory size ({@link Configs.Keys#HEAP_RESERVED_MIN_RATIO}) based on the given configuration.
+   */
+  private int getReservedMemory(@Nullable Map<String, String> config, int defaultValue) {
+    if (config == null) {
+      return defaultValue;
```
anew: similar to getMinHeapRatio

anew:
LGTM

chtyim:
@anew Thanks for the review.

PR #61, opened 2017-08-29
(TWILL-244) Make resource report available to TwillRunnable

chtyim:
- Make resource report available to TwillContext
  - Enhance tracker service to provide more resource related endpoints
    - Provide endpoints to fetch specific part of the report to reduce bandwidth
    - Added a ResourceReporter API to provide programmatic access
  - Make resource report available via TwillContext
- Fixed a concurrent modification bug in ZKDiscoveryService.close


twill-core/src/main/java/org/apache/twill/internal/ResourceReportClient.java
```diff
@@ -0,0 +1,116 @@
...
+  @Override
+  public ResourceReport getResourceReport() {
+    for (URL url : resourceUrls) {
+      try {
+        return fetchURL(url, "", ResourceReport.class);
+      } catch (IOException e) {
+        // Just log a trace as it's ok to not able to fetch resource report
+        LOG.trace("Exception raised when getting resource report from {}.", url, e);
+      }
+    }
```
anwar6953: even if all urls fail to fetch - why is that ok?

chtyim: That's the contract of the `ResourceReporter`. If the report is not available for whatever reason, it'll just return `null` / empty collection.

PR #62, opened 2017-10-10
(TWILL-248) Upgrade to use Netty-4.1

chtyim:
- Also enable ResourceReportClient to use HTTP compression

### DIFF
--- a/twill-yarn/src/main/java/org/apache/twill/yarn/ResourceReportClient.java
+++ b/twill-yarn/src/main/java/org/apache/twill/yarn/ResourceReportClient.java
@@ -17,19 +17,21 @@
  */
 package org.apache.twill.yarn;
 
-import com.google.common.base.Charsets;
-import com.google.common.io.Closeables;
 import org.apache.twill.api.ResourceReport;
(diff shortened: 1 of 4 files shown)


twill-yarn/src/main/java/org/apache/twill/yarn/ResourceReportClient.java
```diff
@@ -52,12 +54,16 @@
   public ResourceReport get() {
     for (URL url : resourceUrls) {
       try {
-        Reader reader = new BufferedReader(new InputStreamReader(url.openStream(), Charsets.UTF_8));
-        try {
+        HttpURLConnection urlConn = (HttpURLConnection) url.openConnection();
```
hsaputra: Is this change related to Netty upgrade?

hsaputra:
One question about change form using BufferedReader,  but looks good.

If passes tests +1

twill-yarn/src/main/java/org/apache/twill/yarn/ResourceReportClient.java
chtyim: Not entirely. This is to enable compression when fetching the report.

PR #65, opened 2018-02-04
(TWILL-254) Update to use ContainerId.fromString

cbaenziger:
This moves away from the deprecated `ConverterUitls.toContainerId` and updates the build to use Hadoop 2.7.2 to avoid folks grabbing master and having a failure trying to run on clusters with code newer than November 18th, 2014 (release of Hadoop 2.6.0).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,7 +171,6 @@
         <junit.version>4.11</junit.version>
         <jopt-simple.version>3.2</jopt-simple.version>
         <commons-compress.version>1.5</commons-compress.version>
-        <hadoop.version>[2.0.2-alpha,2.3.0]</hadoop.version>
         <hadoop20.output.dir>target/hadoop20-classes</hadoop20.output.dir>
     </properties>
 
@@ -538,9 +537,6 @@
             <properties>
                 <hadoop.version>2.3.0</hadoop.version>
             </properties>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
             <build>
                 <plugins>
                     <plugin>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -171,7 +171,7 @@
         <junit.version>4.11</junit.version>
         <jopt-simple.version>3.2</jopt-simple.version>
         <commons-compress.version>1.5</commons-compress.version>
-        <hadoop.version>[2.0.2-alpha,2.3.0]</hadoop.version>
+        <hadoop.version>2.7.2</hadoop.version>
         <hadoop20.output.dir>target/hadoop20-classes</hadoop20.output.dir>
```
serranom: is there any reason to keep this property at all?  seems better to just have it defined in each profile as necessary.

pom.xml
```diff
@@ -680,9 +680,9 @@
             </build>
         </profile>
         <profile>
-            <id>hadoop-2.5</id>
+            <id>hadoop-2.6</id>
             <properties>
-                <hadoop.version>2.5.1</hadoop.version>
+                <hadoop.version>2.6.5</hadoop.version>
```
serranom: why are you eliminating the hadoop-2.5 profile?

twill-yarn/src/main/hadoop22/org/apache/twill/internal/yarn/Hadoop22YarnAMClient.java
```diff
@@ -26,7 +26,7 @@
 /**
  * Wrapper class for AMRMClient for Hadoop version 2.2 or greater.
  */
-public final class Hadoop22YarnAMClient extends Hadoop21YarnAMClient {
+public class Hadoop22YarnAMClient extends Hadoop21YarnAMClient {
```
serranom: why remove final?

twill-yarn/src/main/hadoop20/org/apache/twill/internal/yarn/Hadoop20YarnAMClient.java
```diff
@@ -71,6 +73,11 @@ public Hadoop20YarnAMClient(Configuration conf) {
     this.nmClient = new Hadoop20YarnNMClient(YarnRPC.create(conf), conf);
   }
 
+  @Override
+  private ContainerId containerIdLookup(String containerIdStr) {
+    return (ConverterUtils.toContainerId(containerIdStr));
```
serranom: this should be protected not private

pom.xml
cbaenziger: I didn't see any code specific to 2.5 that 2.3 did not provide?

cbaenziger:
I added a few omissions and corrected the parent  `pom.xml` from my original submission. Now the new code is actually being used and I have updated the default target to Hadoop 2.6 and removed the generic `hadoop.version` as suggested.

twill-yarn/src/main/hadoop20/org/apache/twill/internal/yarn/Hadoop20YarnAMClient.java
cbaenziger: Thank you

twill-yarn/src/main/hadoop22/org/apache/twill/internal/yarn/Hadoop22YarnAMClient.java
cbaenziger: If marked `final`, I can not extend it for [`Hadoop26YarnAMClient`](https://github.com/apache/twill/pull/65/files#diff-8c254de8ae00c6007495979dcb66a986R30)?

cbaenziger:
Thank you for the super fast code review @serranom! I think all comments have been addressed and the code should be complete now. Please let me know if you see anything else necessary. It is great working with Twill! Hopefully I can provide some other ideas as I work with it.

twill-yarn/src/main/hadoop22/org/apache/twill/internal/yarn/Hadoop22YarnAMClient.java
serranom: Ah, okay then.  sorry, I missed that.

pom.xml
chtyim: The profile for different versions of Hadoop is both for version specific code as well as for running unit-tests. It is better to keep it and add the hadoop-2.6

chtyim:
The change looks mostly ok to me, except for the change in the pom.xml profile.

cbaenziger:
Hi @chtyim, thank you for the review. Could you please expand on what you see concerning in the `pom.xml`? For reference, I need to provide a 2.6 target as it is the first version to provide the `ContainerId.fromString()` API.

pom.xml
cbaenziger: @chtyim You are asking me to leave the 2.5 profile alone and add the 2.6 profile?

chtyim: Yes, that's correct.

chtyim:
Changes LGTM.

PR #66, opened 2018-02-28
(TWILL-255) Incorrect logging after memory was adjusted. Does not sho…

yufeldman:
…w memory before adjustment


hsaputra:
+1
LGTM

PR #68, opened 2018-03-19
(TWILL-258) Use loopback address for ZK server

chtyim:
- Fix a race condition in the LocationCacheTest
- Fix race condition in LogLevelChangeTestRun


anew:
Can you change the commit message to mention the race condition fixes?

chtyim:
Updated commit message to include more details about the fixes.

PR #80, opened 2019-01-25
(TWILL-264) Fix Discoverable.hashCode implementation

chtyim:


twill-discovery-api/src/test/java/org/apache/twill/discovery/DiscoverableTest.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.twill.discovery;
```
anew: please add copyright

PR #84, opened 2020-01-15
(TWILL-265) Upgrading kafka version and removing lz4 dependency

CuriousVini:


twill-core/src/test/java/org/apache/twill/kafka/client/KafkaTest.java
```diff
@@ -135,6 +135,9 @@ public void finished() {
           server = new EmbeddedKafkaServer(kafkaServerConfig);
           server.startAndWait();
 
+          // Wait a little while to make sure changes is reflected in broker service
+          TimeUnit.SECONDS.sleep(3);
```
chtyim: It is probably worth to expose the `BrokerService` from the `KafkaClient` and you can have a more efficient wait by use the `BrokerService.addChangeListener` method or simply poll from the `BrokerService.getBrokerList()` method.

CuriousVini: I kept it same as `testBrokerChange()`. But I agree the suggested approach would be a little bit more efficient implementation. 
I will open another pr with that small enhancement to keep the changes in this pr related to kafka version upgrade.

